PR #21, opened 2026-05-13
fix(ilp): harden QWiP dispatcher, disk allocation, and config parsing

mtopolnik:
## Summary

- Switch the QWiP error and connection dispatcher inboxes from drop-newest to drop-oldest under overflow (sf-client.md section 14.6). Watermarks are monotonic, so the most recently observed event carries the most useful state; under bursts the dispatcher now retains the newest entry and discards the oldest. The `getDroppedNotifications()` counter is unchanged.
- macOS `Files.allocate()` no longer masks `F_PREALLOCATE` failures. Real failures such as `ENOSPC` now surface as a `false` return so callers do not end up with a sparse file that `SIGBUS`es on later mmap store. `ENOTSUP`/`EOPNOTSUPP` still falls through to `ftruncate`, so filesystems without `F_PREALLOCATE` continue to work.
- `QwpQueryClient` connect strings now accept both the comma form (`addr=A,B`) and repeated `addr=` keys; both shapes accumulate into a single ordered endpoint list per failover.md section 1. Empty entries continue to be rejected.
- `error_inbox_capacity` minimum is raised from 1 to 16 (sf-client.md section 4.4). Drop-oldest under a bursty error stream needs a wide enough window to preserve the trailing category distribution; values below 16 cannot meet that contract.
- `SegmentRing.rotate()` resets `wakeupRequestedForActive` to `false`. The flag is per-active and was previously set to `true` on rotation, leaving the high-water backup-segment branch a dead path for every active after the first.
- `Files.rename()` brackets each native path pointer in its own `try`/`finally`. An OOM on the second `pathPtr()` previously leaked the first allocation.
- Test cleanup: `QwpRoleRejectCloseRaceTest` brackets `close()` so a throwing close does not leave dangling state, and the dispatcher unit tests are updated to assert the new drop-oldest contract.
- Native libraries are rebuilt to incorporate the `files.c` change.

## Tradeoffs

- Drop-oldest is a visible behavior change. Callers that previously relied on receiving the earliest events under sustained overflow will now see the latest events instead. The dropped counter remains the operational signal that the handler is too slow.
- macOS disk-full conditions that were silently masked may now surface as `Files.allocate() == false`. This is the intended outcome but is observable for any caller that ignored the return value on macOS.
- The new `errorInboxCapacity` floor rejects values in `[1, 15]` at build time. Existing callers using the default (256) or any value >= 16 are unaffected.

## Test plan

- Run the full client test suite: `mvn test`
- New / updated tests that cover the regressions:
  - `SegmentRingTest#testRotationRearmsHighWaterBackupWakeup` for the rotation wakeup flag
  - `QwpQueryClientFromConfigTest` for `addr=` accumulation across comma and repeated-key forms
  - `SenderBuilderErrorApiTest` for the `error_inbox_capacity` minimum of 16
  - `SenderErrorDispatcherTest` and `SenderConnectionDispatcherTest` for the drop-oldest overflow contract and the dropped counter
  - `QwpRoleRejectCloseRaceTest` for the bracketed close on the role-reject race path
- Manual macOS smoke against a near-full APFS volume to confirm `Files.allocate()` returns `false` on `ENOSPC` instead of leaving a sparse file behind.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 22 / 25 (88.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SenderErrorDispatcher.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SenderConnectionDispatcher.java|7|8|87.50%|
|:large_blue_circle:|io/questdb/client/std/Files.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentRing.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpQueryClient.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|2|2|100.00%|

bluestreak01 (approved):

PR #25, opened 2026-05-14
fix(ilp): fix disk-full handling for SF buffer creation

mtopolnik:
## Summary

Splits the responsibilities of `Files.openCleanRW` and `Files.allocate` and unifies the `allocate` contract across Linux, macOS, and Windows.

The two functions previously overlapped: both advanced logical EOF, and the old `allocate` on POSIX simply called `ftruncate` (leaving the file sparse) while on macOS it also over-allocated by `currentSize` on non-empty files. The single production caller pattern -- `openCleanRW(path, sz)` then `allocate(fd, sz)` -- left a sparse file on POSIX and a correctly-reserved one on Windows. When the disk filled later, a producer thread storing into the mmap'd region would trigger a SIGBUS that aborts the JVM (Linux/macOS) or an in-page exception (Windows).

After:

- `openCleanRW(path)` owns the file lifecycle: open RW, truncate to empty. No `size` parameter.
- `allocate(fd, size)` owns "extend EOF and reserve real disk blocks for `[0, target)`." `target = max(size, currentSize)`; the file never shrinks. ENOSPC / `ERROR_DISK_FULL` surface as a clean `false` return. Same observable behaviour on all three platforms.

End-user impact: SF buffer creation hitting a full disk now fails synchronously with a `MmapSegmentException` the sender can recover from, instead of crashing the JVM later.

## Tradeoffs

- API change: `Files.openCleanRW` and `FilesFacade.openCleanRW` lost their `size` parameter. Callers that need a sized file follow with `Files.allocate` (reserves blocks; fails on ENOSPC) or `Files.truncate` (sparse; faster). Two production call sites and several test sites updated.
- On Linux/macOS, if the filesystem rejects `posix_fallocate` / `F_PREALLOCATE`, `allocate` falls back to `ftruncate` and leaves blocks sparse; the SIGBUS risk re-emerges on that filesystem only. Documented on `Files.allocate`'s Javadoc. Windows has no equivalent fallback.
- macOS `allocate` now passes `target - currentSize` to `F_PREALLOCATE` (the correct beyond-EOF length under `F_PEOFPOSMODE`) instead of the full `target`. Behaviour change on non-empty files: stops requesting duplicate allocation for the existing region.

## Test plan

- \`mvn -pl core test\` passes (2219 / 2219, 1 skipped, no failures)
- FilesTest pins the no-shrink and zero-on-fresh-file invariants on \`allocate\`
- MmapSegmentTest covers the full create path including fault-injected \`openCleanRW\` failure and post-\`allocate\` cleanup
- AckWatermarkTest covers the wrong-/missing-size branch that now uses \`openCleanRW + allocate\`


bluestreak01 (approved):
Level 2 review done. Pushed dcc8496 to honour the FilesFacade.allocate contract in AckWatermark.open() (close + unlink, matching MmapSegment.create).

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 8 / 12 (66.67%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/std/DefaultFilesFacade.java|1|2|50.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/AckWatermark.java|4|7|57.14%|
|:large_blue_circle:|io/questdb/client/std/Files.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/MmapSegment.java|1|1|100.00%|

PR #26, opened 2026-05-15
feat(core): add IPv4 and GEOHASH columns to QWP sender

bluestreak01:
Tandem: https://github.com/questdb/questdb/pull/7108

## Summary

This PR bundles several QWP sender improvements landed on the same branch:

### 1. Handshake batch-cap clamping

The QWP WebSocket sender parses the new `X-QWP-Max-Batch-Size` header that the server advertises on the 101 upgrade response, and clamps the configured `auto_flush_bytes` to the smaller of the configured value and the advertised cap. Senders previously had no signal here and would only discover the cap by oversending and reading `STATUS_PARSE_ERROR`.

`auto_flush_bytes=off` (i.e. the user explicitly disabled byte-based auto-flush) stays disabled regardless of the advertised cap — without that branch the clamp would silently re-enable a flush mode the user opted out of.

### 2. IPv4 column on the public Sender

The public `Sender` interface gains `ipv4Column(name, int)` and `ipv4Column(name, String)` overloads. The string form parses dotted-quad via `Numbers.parseIPv4`. Defaults throw `LineSenderException` so unsupported transports (ILP HTTP/TCP/UDP) fail with a clear "current protocol version does not support ipv4" message; only `QwpWebSocketSender` overrides them. Per QuestDB convention the bit pattern 0 (0.0.0.0) is the IPv4 NULL sentinel.

### 3. GEOHASH column on the public Sender

The public `Sender` interface gains `geoHashColumn(name, long bits, int precisionBits)` and `geoHashColumn(name, CharSequence base32)`. The long form masks bits above `precisionBits` so callers may pass an unmasked long without leaking high bits into the top wire byte. The string form decodes the standard geohash base32 alphabet (case-insensitive; reserved chars `a`, `i`, `l`, `o`) via a new `Numbers.parseGeoHashBase32` helper, derives precision as `len * 5`, and caps at 12 chars / 60 bits. Both forms are overridden on `QwpWebSocketSender` and `QwpUdpSender`; the UDP datagram-size estimator gains a `TYPE_GEOHASH` arm in `estimateInProgressColumnPayload` so it stops throwing on geohash columns mid-flush.

### 4. GEOHASH precision persistence fix

`QwpTableBuffer.ColumnBuffer.reset()` (called between batches) used to clear `geohashPrecision` back to `-1` alongside the per-batch value payload. With the strict server-side GEOHASH precision check (see paired oss PR), the next batch's flush iterates the persistent columns schema, finds a GEOHASH column with no values this batch, and the writer falls back to `precision=1`. The server then rejects the batch as `wireBits=1` vs the auto-created `GEOHASH(Nc)`. Precision is now treated as a schema property whose lifetime matches the column type byte. The within-batch reset paths (`truncateTo` and `resetEmptyMetadata`, reached only from `cancelCurrentRow` / in-progress rotation) keep the old reset behaviour so newly-added in-progress columns are still free to acquire a different precision on the next row — pinned by the existing `testCancelRowResetsGeohashPrecisionOnLateAddedColumn`.

### 5. SegmentManagerTest first-spare reorientation

`testProducerWakeupBeatsThePollInterval` was flaking on the coverage CI job. The test's premise turned out to be arithmetically false: with `segSize=120` and `signalAtBytes=90`, one 16-byte append only advances `publishedOffset` to 48 and the high-water wakeup never fires. The test nevertheless passed on healthy machines because the segment manager worker's first iteration (run between `start()` and the assertion, before any `parkNanos`) provisioned the spare via the polling path. JaCoCo coverage instrumentation routinely pushed that first iteration past the 200ms budget. Renamed to `testFirstSpareLandsBeforeFirstPoll`, replaced the misleading comment with one that names the actual mechanism, points readers at `testRotationWakeupTriggersImmediateSparePrep` for the real wakeup contract, and loosened the budgets to 2s / 4s — still strictly below the 5s poll so the "didn't wait for the poll" assertion stands.

### 6. Delta-based `pendingBytes` tracking

`QwpWebSocketSender.sendRow()` no longer walks every entry in `tableBuffers` per row to recompute `totalBufferedBytes()`. It now advances `pendingBytes` by `(current-table-bytes - snapshot)` and re-snaps, with the snapshot anchored on the `table()` switch and reset on the flush / `reset()` paths. Per-row work drops from `O(numTables * numColumns)` to `O(numColumns of the current table)`. Two unit tests in `QwpWebSocketSenderTest` cross-check the running `pendingBytes` against a ground-truth full walk: across a multi-table round-robin (30 rows × 3 tables) and after row cancellation. A JMH micro-bench in the paired oss PR (`QwpTotalBufferedBytesBenchmark`) quantifies the shape.

### 7. Per-row pre-`nextRow` size guard

The existing "row too large for server batch cap" check was gated on `pendingRowCount == 1` and so only inspected the first row of a batch. A small row followed by a huge row would silently slip past the producer-side guard and surface as `ws-close[1009]` on the next operation. The check now runs on every row's column-setter bytes before `nextRow()` commits, leaving prior committed rows in the batch intact and relying on the existing `at()` / `atNow()` `rollbackRow` path to undo the in-progress writes when it throws.

### 8. Flush-time wire-size guard

`flushPendingRows()` refuses to enqueue a message larger than `serverMaxBatchSize`. Schema and dict-delta bytes the encoder adds at message-build time can push a batch of legitimately-sized rows above the wire cap, and users who opted out of byte-based auto-flush can grow batches unbounded. Both cases now throw `LineSenderException` from `flush()` rather than letting an oversize WS frame reach the wire. The guard resets all pending state on the throw (no public API exists to selectively drop the offending portion) and names `droppedRows` in the exception so the caller knows what was lost.

### 9. Strict IPv4 string API

`ipv4Column(name, CharSequence)` is now strict about what counts as a real IPv4 address:

- A `null` reference is a no-op — the column is skipped for the current row and gets null-padded on commit (same shape as `doubleArray(name, null)`).
- The literal string `"null"` (case-insensitive) is rejected. `Numbers.parseIPv4` coerces it to the NULL sentinel; accepting it through this API would let a typo silently round-trip as SQL NULL.
- The dotted-quad `"0.0.0.0"` is rejected for the same reason: its bit pattern is the IPv4 NULL sentinel and the value would silently round-trip as NULL on read.

Javadoc rewritten to document the strict contract. Three unit tests in `QwpWebSocketSenderTest` cover null reference, the four case variants of `"null"`, and `"0.0.0.0"`.

### 10. Closed-sender check precedence on null arguments

Several column setters short-circuit (return `this`) or throw ("cannot be null") on a null argument before delegating to an overload that calls `checkNotClosed()`. On a closed sender the user got a null-flavoured error — or, for the short-circuit setters, a silent no-op — instead of the canonical "Sender is closed". The existing close-flavoured tests all pass non-null arguments, so the null + closed cell of the matrix was never combined.

`checkNotClosed()` now runs at the top of every affected entrypoint, before the null handling. `checkTableSelected()` stays after the null short-circuit so "open + null + no-table" remains a no-op (preserves the `testNullArrayReturnsThis` contract).

Affected entrypoints:

- `QwpWebSocketSender`: `binaryColumn(DirectByteSlice)` (new override of the interface default), `decimalColumn` (4 overloads), `doubleArray` (4 overloads), `geoHashColumn(CharSequence, CharSequence)`, `ipv4Column(CharSequence, CharSequence)`, `longArray` (4 overloads). 15 entrypoints.
- `QwpUdpSender`: same set minus `binaryColumn` (UDP doesn't support BINARY; falls back to the Sender interface default). 13 entrypoints.

Two comprehensive tests (`testNullArgsAfterCloseAllThrowSenderClosed` in each sender's test file) drive every affected entrypoint on a closed sender and assert each throws with "closed" in the message.

### 11. Default client zstd compression level to 1

The library default for `compression_level` (when the user opts into
`compression=zstd` without naming a level) was `3`. Level 3 was burning
CPU for marginal wire savings on typical columnar payloads; level 1 is
the cheapest CPU and the ratio gain at higher levels is usually small
for column-oriented data.

Tests in `QwpQueryClientFromConfigTest` pin the new default
(`testCompressionLevelDefaultIsOneWhenOmitted` and
`testCompressionLevelDefaultIsOneWhenCompressionAlsoOmitted`) via a
new `getCompressionLevelForTest()` accessor — a silent flip back to
3 fails loudly. Existing explicit `compression_level=N` configs are
unchanged.

### 12. Promote initial connect to SYNC when reconnect_* is tuned

`reconnect_max_duration_millis` (and the matching backoff knobs) read
like a general retry budget but actually govern only post-failure
reconnects. A user writing `reconnect_max_duration_millis=120000` with
no explicit `initial_connect_retry` got their budget silently ignored
on the first connect — the canonical footgun (first connect races a
deploy / pod reschedule / load-balancer rotation, the auth-timeout
fires, the sender aborts).

Resolution in `LineSenderBuilder.build()`:
- explicit `initial_connect_retry` / `initial_connect_mode` wins
  (including `off` paired with a non-default reconnect budget — the
  override is opt-out-able);
- otherwise SYNC if any `reconnect_*` knob was set explicitly;
- otherwise OFF (unchanged for plain configs).

Two new tests in `InitialConnectRetryTest` pin the contract:
implicit SYNC under a tuned budget, and explicit `off` suppressing
the implicit promotion.

### 13. Expose server-applied zstd compression level

`QwpQueryClient.getNegotiatedZstdLevel()` and
`WebSocketClient.getServerNegotiatedZstdLevel()` return the level the
server actually applied for the connection, parsed from the echoed
`X-QWP-Content-Encoding` response header. Follows the same shape as
the existing `X-QWP-Version`, `X-QWP-Durable-Ack`, and
`X-QWP-Max-Batch-Size` extractors.

Returns `0` when no zstd is in use — the server picked raw, the
header was absent (older servers), or the value didn't match the
`zstd;level=N` shape. Non-zero values are surfaced as the server
wrote them on the wire (no client-side re-clamp) so a misconfigured
server is observable from user code rather than smoothed over.

Useful for diagnostics (apps logging the effective compression on a
connection) and a prerequisite for the OSS-side e2e test that proves
the server's `qwp.egress.compression.force.level` is enforced on the
wire and reloadable without restart.

## Tradeoffs

- The handshake clamp is a no-op when the header is absent: senders pointed at an older server behave exactly as before.
- The clamp only adjusts the byte budget; it does not touch `auto_flush_rows` or interval-based flush. A user who configured an absurdly large `auto_flush_bytes` against a tiny server cap will see batches flush sooner than they asked for — the desired outcome, but worth knowing.
- IPv4 / GEOHASH sender API is additive: ILP transports get the same default-throwing methods and never have to know about these wire types. Existing client code that didn't call the new methods is unaffected.
- The GEOHASH precision-persistence fix is a behaviour change: a caller who registered a GEOHASH column at precision N in batch 1 and tries precision M ≠ N in batch 2 now gets a client-side `LineSenderException` immediately. Before this fix the divergence reached the wire and the server rejected the whole batch with `DROP_AND_CONTINUE`. Both are errors; the new behaviour fails faster and more clearly, but in-flight code that relied on the cross-batch silence will surface a new exception type.
- The SegmentManagerTest renames a public test method; downstream test scripts that filter by that test name need updating.
- The delta-based `pendingBytes` adds a snapshot field and a few bookkeeping lines in `table()` / `reset()` / `flushPendingRows()`. The per-row hot path is strictly faster, but the table-switch and flush paths each do one extra `getBufferedBytes()` call.
- The per-row pre-`nextRow` guard catches oversize rows N ≥ 2 in a batch that previously surfaced as wire-level disconnects on the next operation. Callers that today depend on the silent acceptance of such rows — there shouldn't be any — will see a synchronous `LineSenderException` from `at()` / `atNow()` instead.
- The flush-time guard drops the entire pending batch on throw rather than preserving any of it. The exception message names `droppedRows` so callers can react, but a caller relying on a partial flush succeeding will see all of it discarded.
- The strict IPv4 string API is a behaviour change for callers that today pass literal `"null"` or `"0.0.0.0"` and expect to store NULL. Both inputs now throw `LineSenderException`. Pass a `null` reference (or omit the setter) to mark the row null; pass any valid dotted-quad to store a real IPv4.
- The closed-sender precedence fix changes the error message for callers that today combine closed sender + null argument: previously they got a null-flavoured message (or a silent no-op for the short-circuit setters); now they get the canonical "Sender is closed". Strictly better; any working-by-accident caller relying on the silent no-op will surface as an exception.
- The `Sender` interface default for `binaryColumn(CharSequence, DirectByteSlice)` still throws "BINARY slice cannot be null" before any closed check on impls that don't override it (HTTP / TCP / UDP — none of which support BINARY anyway, so those paths throw "current protocol does not support binary" regardless). Fixing this at the interface level needs a deeper refactor; the WS impl (the one that supports BINARY) is now consistent.

## Cross-repo

- OSS counterpart: questdb/questdb#7108 — server-side IPv4 ingress, GEOHASH end-to-end fuzz coverage, the strict GEOHASH precision check that motivates the precision-persistence fix here, the cursor sentinel-null fix for IPv4 / UUID / LONG256, and the e2e regression tests pinning the new batch-cap guards.

## Test plan

- mvn -pl core test -Dtest=NumbersTest — happy paths for `parseGeoHashBase32` plus null / empty / >12-char / reserved-char rejection paths.
- mvn -pl core test -Dtest=QwpWebSocketSenderTest — `ipv4Column` and `geoHashColumn` post-close behaviour matches every other column setter; the new delta-tracking unit tests (`testPendingBytesMatchesGroundTruthAcrossTableSwitches`, `testPendingBytesUnchangedByCancelRow`); the strict-IPv4 trio (`testIpv4ColumnStringNullReturnsThis`, `testIpv4ColumnStringRejectsNullLiteral`, `testIpv4ColumnStringRejectsZeroAddress`). 40 tests, 0 failures.
- mvn -pl core test -Dtest=QwpTableBufferTest — `testCancelRowResetsGeohashPrecisionOnLateAddedColumn` still passes (within-batch precision reset preserved); the other 39 tests confirm the cross-batch change doesn't regress anything.
- mvn -pl core test -Dtest=SegmentManagerTest — all 8 tests pass under coverage instrumentation.
- mvn -pl core test -Dtest='%regex[.*[Qq]wp.*]' — full client QWP regression: 1315 tests pass (1310 before this PR + 1 new in `QwpWebSocketSenderTest` + 1 new in `QwpUdpSenderTest` + the existing close-checked entrypoints that now also assert "closed" first).
- End-to-end coverage for the size guards lives in the paired oss PR's `QwpSenderOversizeRowInBatchTest`.


mtopolnik:
#### [PR Coverage check]

:disappointed: **fail** : 101 / 215 (46.98%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|0|34|00.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|2|9|22.22%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|37|102|36.27%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|23|29|79.31%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/OffHeapAppendMemory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|15|16|93.75%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|19|19|100.00%|

mtopolnik:
#### [PR Coverage check]

:disappointed: **fail** : 113 / 244 (46.31%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|0|34|00.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|2|9|22.22%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|49|131|37.40%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|23|29|79.31%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/OffHeapAppendMemory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|15|16|93.75%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|19|19|100.00%|

mtopolnik:
#### [PR Coverage check]

:disappointed: **fail** : 125 / 286 (43.71%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|0|16|00.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|10|48|20.83%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|53|152|34.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|23|29|79.31%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/OffHeapAppendMemory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|15|16|93.75%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|19|19|100.00%|

mtopolnik:
#### [PR Coverage check]

:disappointed: **fail** : 131 / 292 (44.86%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|10|48|20.83%|
|:large_blue_circle:|io/questdb/client/Sender.java|6|22|27.27%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|53|152|34.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|23|29|79.31%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/OffHeapAppendMemory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|15|16|93.75%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|19|19|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 337 / 551 (61.16%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|10|48|20.83%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|53|152|34.87%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|22|42|52.38%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|23|29|79.31%|
|:large_blue_circle:|io/questdb/client/Sender.java|191|240|79.58%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/OffHeapAppendMemory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpQueryClient.java|14|15|93.33%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|19|19|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 337 / 551 (61.16%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|10|48|20.83%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|53|152|34.87%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|22|42|52.38%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|23|29|79.31%|
|:large_blue_circle:|io/questdb/client/Sender.java|191|240|79.58%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/OffHeapAppendMemory.java|5|6|83.33%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpQueryClient.java|14|15|93.33%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|19|19|100.00%|

PR #32, opened 2026-05-23
fix(qwp): fix ingestion stalls after append failures and on sender startup

bluestreak01:
Refs questdb/questdb#7143

This PR bundles two QWP fixes for the upcoming 1.3.1 client release. Both are small, localized, and validated by their own regression tests.

## Microbatch buffer release on append failure

- `QwpWebSocketSender.sealAndSwapBuffer()` left the microbatch buffer in `SENDING` state when `CursorSendEngine.appendBlocking()` threw. No I/O thread ever recycles a buffer the engine never accepted, so the next flush would wait 30 s on the recycle latch and then throw `Timeout waiting for buffer to be recycled`.
- The catch block now puts the buffer back into a non-in-use state on the user thread before rethrowing: `markRecycled()` from `SENDING`, `rollbackSealForRetry()` from `SEALED`.
- The encoded payload itself is dropped, but `flushPendingRows()` aborts its post-enqueue state updates after `sealAndSwapBuffer()` throws, so the source rows in `tableBuffers` and the sent-schema watermark stay intact. The next batch re-emits the same rows together with the full schema and symbol-dict delta — matching the existing "self-sufficient frames" invariant the cursor SF pipeline already relies on.

## Segment manager wakeup on register

- `SegmentManager.register()` now unparks the worker after publishing the new ring. Without the wakeup, if the worker thread acquires the manager lock before `register()` does, it snapshots an empty `rings`, services nothing, and parks for the full poll interval. A ring whose first append does not cross the high-water mark fires no producer-side wakeup either, so the spare never lands until the poll expires.
- The race is pre-existing in the store-and-forward code, latent since the SF feature first landed. It surfaced on the parent-repo CI run only because the mac-other + JaCoCo combination consistently scheduled the worker first; the prior fix (commit 19c5c65) widened the test budget to 2 s but the poll interval is 5 s, so no budget below 5 s could rescue that ordering.
- The fix is one call to `wakeWorker()` after `setManagerWakeup()`. `LockSupport.unpark` is cheap, is a no-op when the worker has not been started, and grants a permit that the next `parkNanos` consumes immediately — both interleavings are covered.

## Tradeoffs

- The microbatch-buffer fix's `close()`-after-failed-flush retries the drain. If the underlying engine is still wedged (e.g. permanent backpressure, like the deliberately-undersized engine in the new test), that retry fails again and surfaces as a separate `LineSenderException`. The test acknowledges this and swallows the close-time rethrow. Real-world transient backpressure is expected to clear between the user's flush call and `close()`, in which case the drain succeeds.
- The segment-manager fix bundles a second unrelated change under a title that names only the microbatch fix. The release-note line for 1.3.1 will undercount the actual scope. Mitigated by clearly separating the two fixes in this description and in the commit history (two separate commits).
- No state-machine extension on either fix: `MicrobatchBuffer.markRecycled()` / `rollbackSealForRetry()` and `SegmentManager.wakeWorker()` already existed.

## Test plan

- New `QwpWebSocketSenderTest.testFlushAppendFailureDoesNotLeaveMicrobatchBufferInUse` (the reproducer from questdb/questdb#7143). Fails on `main` with `buffer0=SENDING, buffer1=FILLING` and the 30 s recycle-timeout suppressed exception; passes in ~80 ms with this change.
- New `SegmentManagerTest.testRegisterAfterWorkerParkedWakesWorker`. Sleeps 250 ms between `start()` and `register()` to guarantee the worker has parked on an empty `rings` snapshot, then asserts the spare lands within 2 s. Without the `wakeWorker()` call this test fails reliably; with it, all 9 `SegmentManagerTest` cases pass.
- Full `QwpWebSocketSenderTest` and the broader QWP client suite still pass.

### DIFF
--- a/core/src/main/java/io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java
+++ b/core/src/main/java/io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java
@@ -3072,6 +3072,18 @@ public class QwpWebSocketSender implements Sender {
             cursorEngine.appendBlocking(toSend.getBufferPtr(), toSend.getBufferPos());
             toSend.markRecycled();
         } catch (Throwable t) {
+            // appendBlocking failed synchronously on the user thread — the
+            // payload never reached the engine, so no I/O thread will
+            // recycle toSend. Recycle it here so a later flush can swap
+            // back to it; flushPendingRows aborts its post-enqueue state
+            // updates after this throw, so the source rows and the
+            // sent-schema watermark stay intact and the next batch re-emits
+            // the same rows along with the full schema + symbol-dict delta.
+            if (toSend.isSending()) {
+                toSend.markRecycled();
+            } else if (toSend.isSealed()) {
+                toSend.rollbackSealForRetry();
+            }
             // Surface any I/O thread error first — appendBlocking itself only
             // throws on PAYLOAD_TOO_LARGE / backpressure deadline, but the
             // I/O loop can have failed independently.
--- a/core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentManager.java
+++ b/core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentManager.java
@@ -247,6 +247,17 @@ public final class SegmentManager implements QuietCloseable {
             }
         }
         ring.setManagerWakeup(this::wakeWorker);
+        // Nudge the worker so it picks up the new ring on its very next
+        // iteration. Without this, register-after-start has a race window:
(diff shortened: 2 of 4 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 3 / 5 (60.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentManager.java|1|1|100.00%|

PR #38, opened 2026-06-03
refactor(qwp): simplify the wire protocol to one version with inline schemas

mtopolnik:
## What

Removes the QWP schema-reference mechanism and the protocol-v2 designation from the Java client. Column schemas now always travel inline on the wire, and the client runs a single protocol version.

This is the client (Phase 2) half of a tandem change; the server half is questdb/questdb#7200. 

## Why

QWP shipped with a schema-reference optimisation: a sender could transmit a column schema once, receive a schema id, and reference that id on later frames. Store-and-forward retry made this unworkable -- a recorded frame can replay against a fresh server connection that never saw the registering frame, leaving the reference dangling. Senders already stopped using reference mode in production, so the wire kept paying for a mechanism nobody used.

## Wire format

Ingress table block (sender -> server):

```
before: name | row_count | col_count | mode | schema_id | [cols if FULL]
after:  name | row_count | col_count | columns (always inline)
```

Egress `RESULT_BATCH` (server -> client): the schema rides only the first batch of a query (`batch_seq == 0`); continuation batches carry rows and bind against the schema parsed there.

The per-frame version byte stays at offset 4, hardcoded to `1` and validated `== 1`. The `X-QWP-Max-Version` handshake header and negotiation machinery stay so a future version bump can re-introduce a range.

## Notable changes

- Decoder: the connection-scoped schema registry is replaced by per-query schema state, invalidated when a new query starts. A continuation batch that arrives before its schema-bearing `batch_seq == 0` is now rejected rather than binding rows to a stale schema.
- API removal (breaking): `Sender.builder().maxSchemasPerConnection(...)` and the `max_schemas_per_connection` connection-string option are gone. Code or config strings that reference them now fail -- with a compile error and an "unknown option" error respectively.
- `SERVER_INFO` is always delivered post-upgrade (no v2 gate), so the failover role filter no longer has a v1 "no role available" branch.

## Trade-offs

- Every query result re-ships its column schema in batch 0, even when the same client re-runs the same query. The old per-connection registry deduped this; the cost is now a few dozen bytes plus a fresh column-name String per column on each batch 0.
- The `maxSchemasPerConnection` builder method disappears; external callers that set it will no longer compile. Acceptable for beta.

## Test plan

- `mvn -pl core test` (client unit + decoder-hardening suites)
- Schema-reference and schema-id-exhaustion tests removed (mechanism gone); other decoder-hardening coverage (truncation, corrupt zstd, oversized varints) retained.
- `SelfSufficientFramesTest` trimmed to its symbol-dict-delta invariant.
- End-to-end multi-batch / wide-schema coverage lives in the server repo's `cutlass-test-suite`, which runs this client against the matching server branch.

## Coordination

Tandem with questdb/questdb#7200. The wire format is breaking; merge the two together.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 31 / 37 (83.78%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpEgressIoThread.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|3|5|60.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpResultBatchDecoder.java|19|22|86.36%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketEncoder.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpColumnWriter.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpQueryClient.java|1|1|100.00%|

mtopolnik:
Superseded by #39, re-created on branch `mt_remove-schema-ref` so the client CI matches the server PR branch name and builds against questdb/questdb#7200 instead of `master`. GitHub auto-closed this PR when the branch was renamed.

PR #40, opened 2026-06-08
fix(qwp): prevent silent row loss on sender reconnect

jerrinot:
This pull request fixes a rare data-loss bug in the QWP store-and-forward sender. When a sender reconnected to the server while your application was still ingesting, it could silently drop the first batch it was about to retransmit and misnumber the rows that followed, leaving them out of the table while `close()` still reported a clean, complete delivery. 

The fix closes the timing window that caused it, so row batches now survive a reconnect-under-load intact. The change is confined to the reconnect repositioning path: steady-state ingestion is untouched, and the only added cost is one extra in-memory segment lookup per reconnect, which is negligible since reconnects are infrequent.

**Implementation details for reviewers**
When the QWP store-and-forward sender reconnects while the application keeps ingesting, it could silently drop the first frame it was about to replay and shift every later frame's sequence number by one. The server then trimmed the unsent span on its next cumulative ACK, so rows went missing even though close() reported a clean delivery.

positionCursorAt() repositions the send cursor after a reconnect, but the producer can publish that frame concurrently - between the cursor's first lookup and its read of the segment's published offset. The offset then sits past the new frame, so the cursor parked one frame too far ahead.

The fix rechecks for the frame right after the published-offset read. That read is the producer's volatile publish barrier, so any newly published frame is now discoverable and the cursor lands exactly on it. If the frame is genuinely not published yet, both lookups miss and the normal send loop picks it up later - no change on the common path.

### DIFF
--- a/core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java
+++ b/core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java
@@ -178,10 +178,18 @@ public final class CursorWebSocketSendLoop implements QuietCloseable {
     // async dispatcher path does NOT set this flag: a user who only watches
     // the async error inbox still gets a loud failure on shutdown.
     private volatile boolean errorSurfacedSynchronously;
-    // fsnAtZero: FSN that wireSeq=0 maps to on the current connection. For
-    // a fresh connection, this is 0. After a reconnect, it's set to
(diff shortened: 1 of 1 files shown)


bluestreak01 (approved):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 4 / 5 (80.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|4|5|80.00%|

PR #42, opened 2026-06-09
fix(qwp): fix missed terminal send errors during Sender close

jerrinot:
Fixes a race where `QwpWebSocketSender.close()` could return successfully even after a terminal QWP/WebSocket error was latched during close, and a related corner where close() could rethrow a wrapped terminal error the user had already caught from `flush()`.

- Track the exact latched instance that `checkError()` has synchronously surfaced and make `close()` suppress only that pre-owned instance, taken as the single read `getSynchronouslySurfacedError()`. This avoids treating a freshly latched terminal as already owned when it appears between two separate latch reads. Guarded by `CloseOwnershipRaceTest` (probabilistic accessor race, near-certain catch per CI run, ~0.15s).
- Simplify the latch so its invariants live in one place: `recordFatal` wraps non-`LineSenderException` causes once at latch time, so every rethrow delivers the same instance — close()'s identity suppression becomes correct in every state, which also fixes a pre-existing corner where `drainOnClose()` re-threw a fresh wrapper for an error the user already caught. `checkUnsurfacedError()` encapsulates the close() safety net and `hasUnsurfacedError()` becomes private, so the racy two-read snapshot is no longer expressible outside the class. The typed `SenderError` payload is derived from the latched `LineSenderServerException` instead of being tracked in a sibling field.
- Rename the latch field `lastError` to `terminalError` (and `getLastError()` to `getTerminalError()`): the field is a write-once, first-writer-wins latch — transient failures reconnect upstream and never reach it — so "last" was misleading.
- `CloseSafetyNetTest` pins both branches of close()'s safety net deterministically: a config-string-only caller who never flushed must see the latched terminal thrown from close(); a caller whose custom error handler already received it must get a silent close.

### DIFF
--- a/.pi/skills/review-pr/SKILL.md
+++ b/.pi/skills/review-pr/SKILL.md
@@ -1,0 +1,339 @@
+---
+name: review-pr
+description: Review a GitHub pull request against QuestDB coding standards. Use when asked to review a PR (by number or URL), optionally with a depth level 0..3. Performs an adversarial, blocking, mission-critical code review covering correctness, concurrency, performance, resource management, tests, and QuestDB conventions, then verifies every finding against source before reporting.
+allowed-tools: bash read subagent
+---
+
+# Review PR
+
+Review the pull request identified by the user's arguments. The arguments (a PR number or URL, optionally with a level token) are appended to this skill as `User: <args>`, or appear in the user's request. Treat that text as `$ARGUMENTS` below.
+
+## Tooling note (pi)
+
+This skill was ported from a Claude Code skill. Map the tools as follows:
+- Reading files → the `read` tool.
+- Searching the repo (the Claude `Grep`/`Glob` steps) → the `bash` tool with `rg`, `grep`, `find`, `ls`. Every "use Grep/Glob" instruction below means "run a real `rg`/`find` search and show it" — do not reason about callsites from memory.
+- `gh` commands → the `bash` tool.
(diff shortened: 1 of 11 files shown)


bluestreak01 (dismissed):

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 42 / 46 (91.30%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|24|28|85.71%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SenderErrorDispatcher.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|11|11|100.00%|

bluestreak01 (approved):
Approved.

Level-3 review (10 reviewers + per-finding source verification, clean `mvn -pl core test-compile`, regression suite green). The error-ownership rework is correct: no correctness, concurrency, resource, or performance blocker found in the production code, and no out-of-diff broken callsites (parent questdb/questdb-enterprise repos traced — zero external callers of the renamed symbols).

The three fixes verified sound, with the identity-stable terminal latch as the backbone:
- Lost-error: close() now takes a single getSynchronouslySurfacedError() read, eliminating the two-read tear.
- M3: drainOnClose(boolean) stops silently on a handler-owned terminal (provably non-hanging).
- C1: per-terminal hasDeliveredTerminalToCustomHandler() replaces the coarse 'any error ever' flag. Identity chain confirmed end to end.

Wire-level terminals always surface loudly (never markTerminal/dispatchError), so the no-loss invariant holds.

Non-blocking follow-ups (nice-to-have, not gating):
- CloseOwnershipRaceTest is effectively a tautology — it never calls checkError() so getSynchronouslySurfacedError() is always null and never executes close(); the lost-error fix would benefit from a real sender-level regression test. (M3 and C1 are genuine red→green tests.)
- Minor: stale hardcoded ref 'QwpWebSocketSender.java:2657' in CloseDrainDoubleSignalTest (actual line 2729); broken {@link #drainOnClose()} -> drainOnClose(boolean) at QwpWebSocketSender.java:1391; hasDeliveredToCustomHandler() now has no production consumer; CloseTerminalConflationTest under-asserts the caught exception.

PR #53, opened 2026-06-17
fix(qwp): robust pooled store-and-forward slot lifecycle — distinct ids, startup recovery, prompt close()

bluestreak01:
## Problem

`SenderPool` builds every sender from one immutable config string. When store-and-forward (SF) is enabled (`sf_dir` set), every sender therefore inherited the **same** `sender_id` (default `"default"`), pointed at the **same** slot dir `<sf_dir>/default`, and took an exclusive `flock` there for its lifetime. The second `borrow()` died on the lock:

```
java.lang.IllegalStateException: sf slot already in use by another process
  [slot=/.../qdb-sf-.../default, holder=pid=...]
  at SlotLock.acquire(SlotLock.java:99)
  at CursorSendEngine.<init>(...)
  at Sender.fromConfig(...)
  at SenderPool.createUnlocked(...)
  at SenderPool.borrow(...)
```

Net effect: **an SF pool could hold at most one live sender** — pooling + SFA was effectively broken. The code even documented the footgun (`Sender.java`: *"multi-sender setups must set this explicitly or the second sender will fail"*) but the pool had no way to satisfy it with a single config string.

Closing the collision exposed a chain of adjacent robustness gaps in the pooled SF slot lifecycle — flock release on teardown, orphan-drain coexistence, recovery of data stranded across restarts, and prompt shutdown — which this PR addresses as one coherent change.

## Fix

When SF is enabled, the pool now hands each sender a **distinct, stable slot id** `<base>-<index>`:
- `<base>` = the configured `sender_id` (default `"default"`), so two pools/processes can share one `sf_dir` by using distinct `sender_id`s.
- `<index>` = a pool slot index in `[0, maxSize)`, recycled **lowest-free-first**.

Because indices are deterministic and reused, a slot dir is re-adopted across restarts and any unacked data it holds is recovered on creation (recovery semantics rely on stable ids — random UUIDs would orphan the data).

**Non-SF paths (HTTP / TCP / memory-mode) are unchanged** — `slotIndex` is `-1` and no slot bookkeeping runs. The cross-writer guard the slot lock exists for is preserved: two pools sharing one `sf_dir` with the **same** base still fail fast.

### Flock-safe slot lifecycle

A slot index is only returned to the free set **after its delegate releases the `flock`**. The pool tracks the in-between states with counters, all folded into the capacity check:

```
all + inFlight + closingSlots + leakedSlots + recoveringSlots < maxSize
```

- **`closingSlots`** — a slot whose delegate is mid-`close()` and still releasing its flock. *Transient*: once the flock drops, the index goes back to the free set. This keeps a concurrent `borrow()` from reclaiming a dir whose lock is still held.
- **`leakedSlots`** — a slot whose delegate `close()` returned with the flock **still held** (the I/O thread refused to stop, observed via `QwpWebSocketSender.isSlotLockReleased()`). `QwpWebSocketSender.close()` has an early-return path (when `cursorSendLoop.close()` throws, it returns before the only flock-release), so "close() returned" does **not** imply "flock released". The index is **retired permanently** — never freed, never reused — so no borrow ever hands out a still-locked dir. Releasing the flock early would be worse (a concurrent borrower could corrupt the dir the zombie I/O thread is still writing). The lost capacity is accounted for in the cap math, so the pool degrades gracefully instead of resurrecting "sf slot already in use". The release/retire logic is shared by the three close-and-reclaim paths (`discardBroken`, `reapIdle`, startup recovery) via one `reclaimSlot()` helper.
- **`recoveringSlots`** — a slot held by the startup-recovery pass (see below); counted so deferred recovery overlapping live borrows can neither over-allocate past `maxSize` nor target a dir being recovered.

A permanent retirement is no longer silent: `SenderPool` has a logger and emits a `WARN` at every retire site naming the slot index and the reduced effective capacity, and exposes the count via a public `leakedSlotCount()` accessor for metrics. Previously the pool had no logger at all, so a pool bleeding capacity degraded to "every `borrow()` times out" with nothing in the logs to explain why.

### Orphan-drain coexistence

The pool co-manages every `<base>-<index>` slot and recovers each slot's unacked data when it (re)creates it. To stop a sibling sender's startup drainer from adopting another **live** pool slot's dir/lock (which would resurface "sf slot already in use"), the pool fences off exactly its own managed slots from orphan draining:
- `LineSenderBuilder.orphanDrainExcludeManagedSlots(String base, int slotCount)` builder option,
- an `OrphanScanner.scan(...)` managed-slot exclusion that excludes only the canonical pool-minted names `<base>-<i>` for `i` in `[0, slotCount)`,
- `build()` passes the configured base + slot count into the orphan scan.

The exclusion is **bounded**, not a coarse `<base>-` prefix: a same-base leftover whose index falls outside the current `[0, maxSize)` range — e.g. `<base>-3` stranded after a `maxSize` shrink across restarts — is treated as a foreign leftover and drained, instead of being silently fenced off and stranded forever. Non-canonical suffixes (`default-007`, `default-foo`) are likewise drained. This is a no-op unless `drain_orphans=on`; foreign leftovers under other names are still drained.

### Startup recovery of stranded managed slots

Because the orphan drainer excludes the pool's live `[0, maxSize)` range, an in-range slot left holding unacked data by a previous run would otherwise be recovered **only** when the pool happens to (re)create that index — so under steady low load it could sit durable-but-undelivered on disk indefinitely. The pool closes this gap by recovering its own stranded managed slots, in two passes:
- **Pass 1** — in-range `[0, maxSize)` slots that hold unacked data and are not already live: reserve the index (under `recoveringSlots`, so no concurrent borrow can target it), build a sender that re-adopts and drains the slot, then close it and free the index. A recoverer whose `close()` leaves the flock held retires the slot permanently, exactly like the other reclaim paths.
- **Pass 2** — same-base slots a **larger** previous run left out of the current range (`<base>-i`, `i >= maxSize`, via `OrphanScanner.listStrandedOutOfRangeManagedSlots`). These are recovered **regardless of `drain_orphans`**, so the default config never strands them.

**Recovery never blocks `build()` and never delays a prompt `close()`:**

- The pooled `QuestDB` handle constructs its pool with `deferStartupRecovery=true`, so recovery does **not** run in the `SenderPool` constructor (against a reachable-but-not-acking server an inline drain would otherwise stall construction for seconds). It is driven instead on the existing `PoolHousekeeper` reap-loop thread — **no new thread**.
- Recovery is a **resumable, per-slot scan**: the housekeeper drives one stranded slot per step (`runStartupRecoveryStep`), back-to-back while work remains (skipping the idle wait so the backlog drains promptly), re-checking its `stop` flag between every step. Direct (non-pooled) `SenderPool` construction still recovers inline to completion (`runStartupRecoveryToCompletion`), single-threaded and unpublished, bounded by one shared acquire-timeout budget so a non-acking server can't pay a per-slot timeout `(maxSize - minSize) ×` over.
- Each per-slot drain is bounded by `RECOVERY_DRAIN_BUDGET_MILLIS` (1s), kept **below** `PoolHousekeeper.STOP_TIMEOUT_MILLIS` (2s). `QuestDBImpl.close()` raises the pool's shutdown signal (`markClosing()`) **before** stopping the housekeeper — a separate `closeStarted` flag guards the actual delegate teardown so the early signal never short-circuits it — so a `close()` landing while a recoverer is mid-drain only ever waits out a single bounded drain. The housekeeper join cannot time out on a fresh slot, the recoverer's flock is released before `close()` returns, and the residual-window "sf slot already in use" on reopen is gone.

*Trade-off:* per-slot recovery drains are capped at 1s each (vs. the old 5s shared budget), trading some per-run recovery headroom on slow links for prompt, race-free shutdown — stranded data stays durable on disk and resumes on the next start.

### Error-safe teardown (run under `-ea`)

`createUnlocked()` runs a heavy native build path (mmap, flock, WebSocket connect) and delegate `close()`/`flush()` can throw an `Error` (e.g. an `-ea` `AssertionError`, `OutOfMemoryError`), not just a `RuntimeException`. Every best-effort teardown and creation site is therefore `catch (Throwable)` so a stray `Error` can't strand sibling flocks, over-count slot bookkeeping, kill the housekeeper daemon, or leak a half-built pool:
- `SenderPool` / `QueryClientPool` prewarm, `borrow()` / `acquire()`, `createUnlocked()`, and `discardBroken` / `reapIdle` slot accounting (moved into `finally` so it always runs).
- `PooledSender.close()` tracks normal `flush()` completion (`flushed=true`) and routes any abnormal exit — `RuntimeException` **or** `Error` — through `discardBroken`, so a sender that failed to flush is never recycled with unsent rows.
- `QuestDBImpl` orchestration: the ctor cleanup catch is `Throwable`, and each teardown step (`housekeeper.stop()`, `queryPool.close()`, `senderPool.close()`) is independently guarded via `closeQuietly` helpers so a Throwable from an earlier step can't skip closing the `SenderPool`, which runs last and owns the flock/mmap/I-O-thread resources.

### Construction-race tolerance

The pool builds senders outside its lock, so two first-borrows sharing one `sf_dir` can race into `mkdir(sf_dir)`; the loser previously failed with a spurious `could not create sf_dir`. `build()` now treats a benign creation race (the dir exists afterwards) as success.

## ⚠️ Build-target change: minimum JDK lowered from 11 to 8

> **Heads-up for reviewers / downstream:** this PR also lowers the client's
> minimum supported JDK from **Java 11 to Java 8** and adds the CI to enforce
> it. This is independent of the SF-lifecycle work above and has a wider blast
> radius (it touches a numeric hot path used by *every* `double` serialization).
> A reviewer (M-B) reasonably recommended splitting it into its own PR; it is
> documented here rather than split. Read this section before approving.

### What changes for consumers

- The artifact `io.questdb:questdb-client` now **compiles and runs on Java 8+**
  (previously Java 11+). **Java 8 is the release source of truth** — the
  artifact is built and released from JDK 8.
- No source-compatible API was removed; this is purely a *lowering* of the
  minimum runtime/build JDK. Existing Java 11+ consumers are unaffected.
- The bytecode target is JDK-driven (see below): building on JDK 8 emits
  `1.8` bytecode; building on JDK 11+ emits `${javac.target}` (default `11`).

### How the multi-JDK build works

- **Two auto-activated Maven profiles** select everything by the JDK running the
  build — no flag needed:
  - `java8` (activated on JDK 1.8): `source/target=1.8`, excludes
    `module-info.java`, drops the `--add-exports` flags, logback `1.3.x`.
  - `java11+` (activated on JDK ≥ 11): `source/target=${javac.target}`
    (default 11), keeps `--add-exports java.base/jdk.internal.math`, logback
    `1.5.x`.
- **Per-JDK source roots** `src/main/java8` and `src/main/java11`, added to the
  compile path by `build-helper-maven-plugin` via the `compat.src.dir`
  property. Each holds a JDK-specific copy of two tiny bridge classes:
  - **`Compat`** — shims APIs missing on Java 8 (e.g. `Thread.onSpinWait()`
    becomes a no-op; PID lookup uses the Java 8 path).
  - **`FdBig`** — bridge over the JDK's internal `FDBigInteger` used by the
    `double`→string fast path: the java8 copy targets `sun.misc.FDBigInteger`,
    the java11 copy targets `jdk.internal.math.FDBigInteger` (reached via the
    compile-time `--add-exports`).
- **`Numbers`** double-to-string path was rerouted from a direct
  `jdk.internal.math.FDBigInteger` reference to the `FdBig` bridge so it
  compiles on both JDKs. **This is the high-blast-radius change** — it sits on
  the hot path for every serialized `double`. Behaviour is unchanged; only the
  class it routes through differs per JDK.
- **`Unsafe.addExports` removed** (along with `JAVA_BASE_MODULE` /
  `implAddExports`): the old *runtime* reflective self-injection of
  `jdk.internal.math` is gone. On JDK 11+ the export is now a **compile-time**
  flag in the `java11+` profile; on JDK 8 it is not needed at all.
- **logback** is test-scope only and downgraded to the `1.3.x` line on Java 8
  (`1.5.x` requires Java 11).

### New CI (`.github/workflows/ci.yml`)

- **JDK 8** — `clean install` with `-P javadoc`: compiles, runs the **full**
  test suite against the committed native libs, and builds the javadoc jar.
  This is the source-of-truth job.
- **JDK 25** — compile-only smoke (`-DskipTests clean package -P javadoc`) of
  main + test sources and the javadoc jar, to guard against breakage when the
  client is consumed as a submodule of the JDK-25 `questdb` parent repo.

### Also bundled (flagged by M-B, listed for completeness)

- `QueryClientPool` `startHook` NPE fix on the prewarm-from-production path.
- Throwable-teardown hardening across the pool teardown paths (also load-bearing
  for the SF lifecycle work above).

## Changes

- **`Sender.LineSenderBuilder`** — additive introspection accessors `isStoreAndForwardEnabled()`, `getConfiguredSenderId()`, `getConfiguredSfDir()`; the `orphanDrainExcludeManagedSlots(base, slotCount)` option; `build()` wires the managed base + slot count into the orphan scan and tolerates a benign concurrent `mkdir(sf_dir)` race. The pool probes the config once at construction, so a bad config fails eagerly even when `minSize == 0`.
- **`QwpWebSocketSender`** — `isSlotLockReleased()` (set true only after the flock-release block) so the pool can tell whether `close()` actually dropped the flock.
- **`OrphanScanner`** — bounded managed-slot exclusion (canonical `<base>-<i>` for `i` in `[0, slotCount)` only); `listStrandedOutOfRangeManagedSlots()` backing pass-2 recovery.
- **`PooledSender`** — stable `slotIndex` (`-1` when SF off); `close()` discards on any abnormal `flush()` exit.
- **`SenderPool`** — per-slot `sender_id` derivation; slot-index allocator; flock-safe lifecycle via `closingSlots` / `leakedSlots` / `reclaimSlot()`; bounded managed-slot orphan-drain fencing; resumable per-slot startup recovery (`recoverOneSlotStep`, `runStartupRecoveryStep`, `runStartupRecoveryToCompletion`) with `recoveringSlots` cap accounting; `markClosing()` + `closeStarted` for prompt, cancellable shutdown; Throwable-safe teardown; logger + `WARN` on permanent retirement + `leakedSlotCount()` accessor.
- **`PoolHousekeeper`** — drives recovery one slot per loop iteration on its reap thread, re-checking `stop` between steps; named `STOP_TIMEOUT_MILLIS`; Throwable-safe reap loop.
- **`QuestDBImpl`** — constructs the sender pool with `deferStartupRecovery=true`; signals `markClosing()` before stopping the housekeeper; independently guarded teardown steps; Throwable ctor cleanup catch. Package-private seam constructor (senderFactory + connectHook) for white-box error-injection tests.
- **`QueryClientPool` / `QueryWorker`** — Throwable-safe teardown / prewarm / acquire; `QueryClientPool` implements `AutoCloseable`.
- **`QuestDBBuilder.build()`** — documents that startup recovery is deferred and `build()` does not block on it.
- **Build / JDK 8 (independent of the SF work — see the build-target section above)** — `core/pom.xml` gains the auto-activated `java8` / `java11+` profiles, `build-helper-maven-plugin` (per-JDK source roots via `compat.src.dir`), JDK-driven `source/target`, and a profile-scoped `logback.version`; new `src/main/java8` + `src/main/java11` `Compat` / `FdBig` bridges; `Numbers` reroutes the `double`→string path through `FdBig`; `Unsafe` drops `addExports` / `JAVA_BASE_MODULE` / `implAddExports` (the `jdk.internal.math` export is now compile-time, JDK 11+ only); new `.github/workflows/ci.yml` (JDK 8 build+test+javadoc, JDK 25 compile smoke).

## Tests

New / expanded white-box coverage (each behavioural guard verified red→green by neutering the fix):

- **`SenderPoolSfTest`** (27) — two concurrent SF senders (original repro); grow-to-max coexistence + extra borrow times out; configured-base slot naming; slot reuse and reap-and-reuse index recycling; repeated-saturation invariant; end-to-end ingest; cross-pool flock guard; distinct-base sharing; `close()` releases all slots; **leaked slot retired permanently** when delegate `close()` leaves the flock held (via `discardBroken` **and** `reapIdle`); leaked slot **observable** (WARN + `leakedSlotCount()`); recovery replay through a re-adopted slot; concurrent borrow/return stress; `drain_orphans` does not cannibalize sibling slots but still drains foreign orphans; shrinking `maxSize` drains stranded **out-of-range** slots; in-range idle slot recovered at startup under steady low load; default config recovers out-of-range slots after shrink; startup recovery **bounded by a shared budget**; **deferred construction does not block** + **delivers when driven** (idempotent); retire-on-recovery when the recoverer keeps the flock; benign `sf_dir` `mkdir` race (concurrent + deterministic pre-existing-dir); **`close()` mid-recovery stops building on a closing pool** (the C1 race — fake recoverer parks a drain, shutdown signalled mid-drain, next slot must not be built).
- **`SenderPoolTest`** (19) — non-SF coverage incl. Error-safe reap/close teardown wrapped in `assertMemoryLeak`; borrow-path SF slot-index release on creation failure; flush-`Error` → `discardBroken`.
- **`SenderPoolErrorSafetyTest`** (3) / **`QuestDBImplErrorSafetyTest`** (1) / **`QueryClientPoolErrorSafetyTest`** (5) — `Throwable` (not `RuntimeException`) on every creation/teardown path, via Proxy/`connectHook` seams.
- **`OrphanScannerTest`** (14) — bounded in-range exclusion, out-of-range drain after shrink, non-canonical names drained, `isManagedSlot` matrix.
- **`SlotLockReleasedContractTest`** (2) — real `QwpWebSocketSender.close()` paths: clean close ⇒ `isSlotLockReleased()==true`; forced `!ioThreadStopped` early-return ⇒ `false`.

## Verification

- `SenderPoolSfTest` 27/27 ✅ · `SenderPoolTest` 19/19 ✅ · `SenderPoolErrorSafetyTest` 3/3 ✅ · `QuestDBImplErrorSafetyTest` 1/1 ✅ · `QueryClientPoolErrorSafetyTest` 5/5 ✅ · `OrphanScannerTest` 14/14 ✅ · `SlotLockReleasedContractTest` 2/2 ✅
- `QuestDBBuilderTest` / `LineSenderBuilderTest` and the WS lifecycle / `sf.**` / `sf.cursor.**` suites remain green (no regression).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,62 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  # JDK 8 is the source of truth: the client ships as a Java 8 artifact
+  # (io.questdb:questdb-client) and is released from JDK 8, so on JDK 8 it must
+  # compile, the full test suite must pass against the committed native
+  # libraries, and the javadoc jar must build (-P javadoc attaches it at the
+  # package phase). The committed native .so/.dylib/.dll are enough -- the only
+  # git submodule (zstd) is needed solely for C++ native rebuilds, not here.
+  build-jdk8:
+    name: Build, test & javadoc (JDK 8)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "8"
+          cache: maven
+
+      - name: Compile, test, and build javadoc
+        run: mvn -B -ntp -P javadoc clean install
+
+  # The client is also consumed as a submodule of the main questdb repo, which
+  # builds on JDK 25. Guard against JDK 25 compile breakage here (main + test
+  # sources, both modules), but do NOT run the tests -- the parent repo runs
+  # them against a real server.
+  compile-jdk25:
+    name: Compile smoke (JDK 25)
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Check out
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 25
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "25"
+          cache: maven
+
+      - name: Compile main and test sources (no tests run)
+        run: mvn -B -ntp -DskipTests clean test-compile
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,11 +40,12 @@ jobs:
         run: mvn -B -ntp -P javadoc clean install
 
   # The client is also consumed as a submodule of the main questdb repo, which
-  # builds on JDK 25. Guard against JDK 25 compile breakage here (main + test
-  # sources, both modules), but do NOT run the tests -- the parent repo runs
-  # them against a real server.
+  # builds on JDK 25. Guard against JDK 25 compile breakage (main + test
+  # sources, both modules) and confirm the javadoc jar builds on JDK 25 too
+  # (-P javadoc attaches it at the package phase). Do NOT run the tests -- the
+  # parent repo runs them against a real server.
   compile-jdk25:
-    name: Compile smoke (JDK 25)
+    name: Compile & javadoc smoke (JDK 25)
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
@@ -58,5 +59,5 @@ jobs:
           java-version: "25"
           cache: maven
 
-      - name: Compile main and test sources (no tests run)
-        run: mvn -B -ntp -DskipTests clean test-compile
+      - name: Compile (main + test) and build javadoc (no tests run)
+        run: mvn -B -ntp -P javadoc -DskipTests clean package
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -7,9 +7,30 @@ Guidance for Claude Code when working with the QuestDB Java client
 
 The standalone client repo for QuestDB ingestion (legacy ILP and QWP). It is
 a Git submodule of the `questdb` repository and ships as the
-`io.questdb:questdb-client` Maven artifact. The module targets Java 11, so
-only legacy Java features are available here — no enhanced switch, no
-multiline strings, no pattern variables in `instanceof`.
+`io.questdb:questdb-client` Maven artifact.
+
+**Dual JDK target — always aim at both.** The client ships as a Java 8
+artifact (JDK 8 is the source of truth: CI builds, tests, and releases on
+JDK 8) but must also compile cleanly on JDK 11+ (CI runs a JDK 25
+compile/javadoc smoke job). The right Maven profile (`java8` /
+`java11+`) auto-activates from the running JDK; a `src/main/java8` vs
+`src/main/java11` source root supplies the per-JDK shims. Practically this
+means **Java 8 is the language/API floor** — every change must build on
+both, so write to Java 8 and never introduce an API or syntax that only a
+newer JDK has.
+
+Disallowed here (newer than Java 8), among others:
+- `var` local-variable type inference (10), enhanced `switch` (14),
+  multiline text blocks (15), pattern variables in `instanceof` (16)
+- `List.of` / `List.copyOf` / `Map.of` and friends (9/10),
+  `String.repeat` / `String.strip` / `String.isBlank` (11)
+- `ProcessHandle` (9), `Thread.onSpinWait` (9),
+  the `Provider(String, String, String)` ctor (9 — use the Java 8
+  `Provider(String, double, String)` overload)
+
+When in doubt, check the API's `@since` and keep it `<= 1.8`. JDK
+9+-only behaviour belongs behind the `src/main/java11` shim with a Java 8
+counterpart in `src/main/java8`.
 
 ## Git & PR Conventions
 
@@ -35,12 +56,17 @@ multiline strings, no pattern variables in `instanceof`.
 
 ## Build
 
-Standard Maven build. Targets Java 11:
+Standard Maven build. Dual-target (Java 8 floor, Java 11+ also supported);
+the `java8` / `java11+` profile auto-activates from the JDK running Maven:
 
 ```bash
 mvn clean install -DskipTests   # build + install to local Maven cache
 mvn -pl core test                # run client unit tests
 ```
+
+Build on a JDK 8 to validate the shipping artifact (the source-of-truth
+target); also confirm it compiles on a modern JDK (11+) before merging, the
+same two fronts CI guards.
 
 The parent `questdb` repo's `local-client` profile pulls this module as a
 sub-module so server changes can build against unpublished client code; if
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -37,6 +37,10 @@
         <test.exclude>None</test.exclude>
         <test.include>%regex[.*[^o].class]</test.include><!-- exclude module-info.class-->
         <jmh.version>1.37</jmh.version>
+        <!-- JDK-version-specific source root; overridden in the java8 profile -->
+        <compat.src.dir>src/main/java11</compat.src.dir>
+        <!-- logback 1.5.x needs Java 11; the java8 profile drops to the 1.3.x line -->
+        <logback.version>1.5.25</logback.version>
     </properties>
 
     <version>1.3.5-SNAPSHOT</version>
@@ -184,6 +188,27 @@
                     </excludes>
                 </configuration>
             </plugin>
+            <!-- Adds the JDK-version-specific source root (src/main/java11 or src/main/java8)
+                 so the FDBigInteger bridge / module-access shim compiles on each JDK. -->
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>3.6.0</version>
+                <executions>
+                    <execution>
+                        <id>add-compat-source</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>${compat.src.dir}</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
@@ -206,7 +231,7 @@
                         </executions>
                         <configuration>
                             <doclint>none</doclint>
-                            <source>11</source>
+                            <source>1.8</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
                             <additionalJOptions>
                                 <additionalJOption>${compilerArg1}</additionalJOption>
@@ -338,7 +363,7 @@
                         </executions>
                         <configuration>
                             <doclint>none</doclint>
-                            <source>11</source>
+                            <source>1.8</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
                             <additionalJOptions>
                                 <additionalJOption>${compilerArg1}</additionalJOption>
@@ -466,6 +491,43 @@
                 </dependency>
             </dependencies>
         </profile>
+        <profile>
+            <id>java8</id>
+            <properties>
+                <jdk.version>8</jdk.version>
+                <java.enforce.version>[1.8,11)</java.enforce.version>
+                <questdb.artifactid>questdb</questdb.artifactid>
+                <!-- JDK8 javac rejects the add-exports flag; use a harmless no-op flag instead -->
+                <compilerArg1>-Xlint:none</compilerArg1>
+                <compilerArg2>-Xlint:none</compilerArg2>
+                <!-- module-info.java cannot be compiled on JDK8; exclude it from both the main
+                     (<excludes> = excludeTestPattern1) and test (<testExcludes> = excludePattern1)
+                     source sets. -->
+                <excludePattern1>**/module-info.java</excludePattern1>
+                <excludeTestPattern1>**/module-info.java</excludeTestPattern1>
+                <compat.src.dir>src/main/java8</compat.src.dir>
+                <javac.compile.source>1.8</javac.compile.source>
+                <javac.compile.target>1.8</javac.compile.target>
+                <!-- logback 1.5.x requires Java 11; the 1.3.x line is the Java 8 compatible branch -->
+                <logback.version>1.3.15</logback.version>
+            </properties>
+            <activation>
+                <jdk>1.8</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.jetbrains</groupId>
+                    <artifactId>annotations</artifactId>
+                    <version>17.0.0</version>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                    <version>2.0.17</version>
+                </dependency>
+            </dependencies>
+        </profile>
     </profiles>
 
     <dependencies>
@@ -485,7 +547,7 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.5.25</version>
+            <version>${logback.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -231,7 +231,7 @@
                         </executions>
                         <configuration>
                             <doclint>none</doclint>
-                            <source>1.8</source>
+                            <source>${javac.compile.source}</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
                             <additionalJOptions>
                                 <additionalJOption>${compilerArg1}</additionalJOption>
@@ -363,7 +363,7 @@
                         </executions>
                         <configuration>
                             <doclint>none</doclint>
-                            <source>1.8</source>
+                            <source>${javac.compile.source}</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
                             <additionalJOptions>
                                 <additionalJOption>${compilerArg1}</additionalJOption>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -234,8 +234,8 @@
                             <source>${javac.compile.source}</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
                             <additionalJOptions>
-                                <additionalJOption>${compilerArg1}</additionalJOption>
-                                <additionalJOption>${compilerArg2}</additionalJOption>
+                                <additionalJOption>${javadocJOption1}</additionalJOption>
+                                <additionalJOption>${javadocJOption2}</additionalJOption>
                             </additionalJOptions>
                             <sourceFileExcludes>
                                 <sourceFileExclude>${excludePattern1}</sourceFileExclude>
@@ -366,8 +366,8 @@
                             <source>${javac.compile.source}</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
                             <additionalJOptions>
-                                <additionalJOption>${compilerArg1}</additionalJOption>
-                                <additionalJOption>${compilerArg2}</additionalJOption>
+                                <additionalJOption>${javadocJOption1}</additionalJOption>
+                                <additionalJOption>${javadocJOption2}</additionalJOption>
                             </additionalJOptions>
                             <sourceFileExcludes>
                                 <sourceFileExclude>${excludePattern1}</sourceFileExclude>
@@ -469,6 +469,11 @@
                 <questdb.artifactid>questdb</questdb.artifactid>
                 <compilerArg1>--add-exports</compilerArg1>
                 <compilerArg2>java.base/jdk.internal.math=io.questdb.client</compilerArg2>
+                <!-- javadoc needs the same module export to resolve jdk.internal.math.FDBigInteger
+                     referenced by the src/main/java11 FdBig bridge. Kept separate from the javac
+                     compilerArgs because javadoc and javac accept different option sets. -->
+                <javadocJOption1>--add-exports</javadocJOption1>
+                <javadocJOption2>java.base/jdk.internal.math=io.questdb.client</javadocJOption2>
                 <excludePattern1>nothing-to-exclude-dummy-value-include-all-java11plus</excludePattern1>
                 <excludeTestPattern1>nothing-to-exclude-dummy-value-include-all-java11plus</excludeTestPattern1>
                 <javac.compile.source>${javac.target}</javac.compile.source>
@@ -500,6 +505,10 @@
                 <!-- JDK8 javac rejects the add-exports flag; use a harmless no-op flag instead -->
                 <compilerArg1>-Xlint:none</compilerArg1>
                 <compilerArg2>-Xlint:none</compilerArg2>
+                <!-- JDK8 javadoc rejects javac's -Xlint and needs no module export (the java8 FdBig
+                     bridge uses the open sun.misc.FDBigInteger). Pass a harmless no-op instead. -->
+                <javadocJOption1>-quiet</javadocJOption1>
+                <javadocJOption2>-quiet</javadocJOption2>
                 <!-- module-info.java cannot be compiled on JDK8; exclude it from both the main
                      (<excludes> = excludeTestPattern1) and test (<testExcludes> = excludePattern1)
                      source sets. -->
--- a/core/src/main/java/io/questdb/client/QuestDBBuilder.java
+++ b/core/src/main/java/io/questdb/client/QuestDBBuilder.java
@@ -73,6 +73,13 @@ public final class QuestDBBuilder {
      * connections in each pool; further slots are allocated lazily up to
      * {@code max} when load demands and reaped back to {@code min} when
      * idle.
+     * <p>
+     * Non-blocking on startup recovery: when store-and-forward is enabled,
+     * unacked data a previous run left in this pool's managed slots is
+     * recovered on a background housekeeper thread shortly after this method
+     * returns -- so {@code build()} does not block on a slow or
+     * reachable-but-not-acking server. The recovered data is durable on disk
+     * and is delivered once the server acks; until then it stays preserved.
      */
     public QuestDB build() {
         if (ingestConfig == null) {
--- a/core/src/main/java/io/questdb/client/Sender.java
+++ b/core/src/main/java/io/questdb/client/Sender.java
@@ -1478,7 +1478,15 @@ public interface Sender extends Closeable, ArraySender<Sender> {
                 } else {
                     if (!Files.exists(sfDir)) {
                         int rc = Files.mkdir(sfDir, Files.DIR_MODE_DEFAULT);
-                        if (rc != 0) {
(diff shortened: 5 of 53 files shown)


bluestreak01:
## Review (level 3, mission-critical pass)

Read all changed source in full, built the change-surface map with real searches, traced the cap-check invariant and all reclaim/recovery paths, compiled the module, and ran all 72 new/changed tests (green).

### Change surface (summary)

- **`SenderPool`** (+808): per-slot `sender_id` derivation, slot-index allocator, `closingSlots`/`leakedSlots`/`recoveringSlots` accounting, `reclaimSlot()`, resumable startup recovery, `markClosing()`+`closeStarted`, Throwable-safe teardown, logger + `leakedSlotCount()`.
- **`PooledSender`**: stable `slotIndex`; `close()` discards on any abnormal `flush()` exit.
- **`QwpWebSocketSender`**: `slotLockReleased` + `isSlotLockReleased()`.
- **`OrphanScanner`**: bounded managed-slot exclusion + `listStrandedOutOfRangeManagedSlots()` + `parseCanonicalIndex()`.
- **`PoolHousekeeper`**: drives recovery one slot/iteration, `STOP_TIMEOUT_MILLIS`, Throwable-safe.
- **`QuestDBImpl`**: deferred recovery, `markClosing()` before housekeeper stop, guarded teardown.
- **`QueryClientPool`/`QueryWorker`**: Throwable-widened teardown; `AutoCloseable`.
- **`Sender.LineSenderBuilder`**: introspection accessors + `orphanDrainExcludeManagedSlots`; benign `mkdir` race tolerance.

Callsite inventory confirms the new symbols are reached only from `SenderPool`, `QuestDBImpl`, `PoolHousekeeper`, `Sender.build()`, and tests — blast radius is contained within the pool subsystem.

### Critical

None. I specifically tried to break the central invariant and the recovery/close races and could not.

The cap-check invariant is airtight: every `slotInUse[i]==true` entry is counted by exactly one of `{all-membership, inFlightCreations, closingSlots, leakedSlots, recoveringSlots}`, and every counter increment is paired with a `slotInUse` write under `lock`. Therefore `sum < maxSize ⟺ a free index exists`, so `allocateSlotIndex()`'s defensive throw is genuinely unreachable in production, and concurrent borrow vs. deferred recovery can neither over-allocate nor target the same dir.

### Moderate

**M1 — No test exercises the production SF-recovery path (real `PoolHousekeeper` + `QuestDBImpl.close()` ordering) — out-of-diff coverage gap**
- `PoolHousekeeper.java` recovery-driving lines are **0% covered** (per the coverage bot); `QuestDBImpl.close()` is 45%.
- The recovery *logic* is well covered, but only via two non-production drivers: the inline constructor path (raw `SenderPool`, `deferStartupRecovery=false`) and hand-driven reflection calls. The C1 test (`testCloseDuringDeferredRecoveryStopsBuildingOnClosingPool`) **mimics** the housekeeper with a hand-rolled `Thread` that calls `runStartupRecoveryStep` in a loop — it does **not** run the real `PoolHousekeeper.runLoop()`.
- So the actual production wiring for a pooled `QuestDB` handle is untested end-to-end: `runLoop()` calling `runStartupRecoveryStep()`, the "skip idle wait while `recovering`" branch, the `stop` re-check between steps, the Throwable swallow, and `QuestDBImpl.close()` raising `markClosing()` **before** `housekeeper.stop()` with the join not timing out on a fresh slot. This is exactly the "prompt, race-free close()" headline guarantee.
- **Suggested fix:** add one integration test that builds a real `QuestDB` handle (or `QuestDBImpl` via the package-private seam) against a silent server with seeded stranded slots, lets the real housekeeper drain them, and asserts `close()` returns within `STOP_TIMEOUT_MILLIS` and the data is recovered against an acking server.

### Minor

- **m1 — Internal pool concepts leak into the public `LineSenderBuilder` API.** `LineSenderBuilder` is a nested type of the `Sender` interface, so it is implicitly `public static final`; the new `getConfiguredSenderId()`/`getConfiguredSfDir()`/`isStoreAndForwardEnabled()`/`orphanDrainExcludeManagedSlots(base, slotCount)` are now frozen public API despite being documented as internal hooks. `orphanDrainExcludeManagedSlots` in particular exposes a pool-internal slot-namespace concept to every user. Consider a non-public seam. (Not blocking — no cleaner cross-package mechanism without reflection.)
- **m2 — Test-count claims off by one.** The body says `SenderPoolSfTest (27)` / `27/27 ✅`, but the suite has **28** `@Test` methods and the run reports `Tests run: 28`.
- **m3 — No `Fixes #NNN` in the body.** Per repo convention, link the tracked issue at the top if one exists (label is `bug`).
- **m4 — Inline-path recovery budget coupled to `acquireTimeoutMillis`.** `runStartupRecoveryToCompletion()` caps the whole inline scan at one `acquireTimeoutMillis`; a direct `SenderPool` built with a tiny acquire timeout effectively skips startup recovery (deferred to later borrow re-adoption). Only affects direct `SenderPool` usage — `QuestDBImpl` uses the deferred housekeeper path bounded by `RECOVERY_DRAIN_BUDGET_MILLIS` — and is a documented tradeoff, but easy to misread.

### Downgraded (verified false positives)

- **Non-volatile `slotLockReleased` visibility bug** — FALSE. Every write (`QwpWebSocketSender.close()`) and read (`flockReleased()` via `isSlotLockReleased()`) is on the *same thread* in all three reclaim paths.
- **`allocateSlotIndex()` can throw under concurrent borrow/recovery** — FALSE. The `sum == popcount(slotInUse)` invariant makes the throw unreachable.
- **`borrow()` success path leaks a slot if `all.add()` throws OOM** — DISMISSED (unreachable): `all` is pre-sized to `maxSize`, never resizes.
- **`markClosing()` doesn't signal `slotReleased`, stranding waiters** — FALSE. Always followed by `close()` (`signalAll`); waiters time out and re-check `closed` regardless.
- **OFF-mode recovery delegate can't reconnect mid-drain → data loss** — DISMISSED. Data stays durable on disk for a later attempt; recovery is best-effort and bounded.
- **`listStrandedOutOfRangeManagedSlots` with `managedSlotCount==0` lists everything** — DISMISSED (unreachable): the only production caller passes `maxSize >= 1`.

### Summary

**Verdict: approve** — with M1 (production-path test coverage) worth addressing before merge for a mission-critical close path; minors at the author's discretion.

- The original collision bug is real and the fix is correct: distinct `<base>-<index>` slot ids, verified by `testTwoConcurrentSfSendersGetDistinctSlots` and the full suite (72/72 green on a clean local build). The flock-safe lifecycle, Throwable-widened teardown, and deferred/cancellable recovery are carefully reasoned and the central capacity invariant holds under concurrency.
- Draft findings: 11 considered → 5 verified (1 Moderate, 4 Minor), 6 dropped as false positives/unreachable.
- In-diff vs out-of-diff: 4 in-diff, 1 out-of-diff (M1 — the untested production `PoolHousekeeper`/`QuestDBImpl.close()` integration). No regressions found; the main tradeoff (per-slot 1s recovery drain vs. the old 5s shared budget, plus the documented black-holed-connect residual window on close) is sound and honestly documented in-code.

<sub>🤖 Generated with review-pr (level 3)</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 313 / 385 (81.30%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/impl/PoolHousekeeper.java|0|13|00.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|0|1|00.00%|
|:large_blue_circle:|io/questdb/client/cutlass/line/http/LineHttpSenderV3.java|0|4|00.00%|
|:large_blue_circle:|io/questdb/client/impl/QuestDBImpl.java|10|22|45.45%|
|:large_blue_circle:|io/questdb/client/impl/QueryWorker.java|6|10|60.00%|
|:large_blue_circle:|io/questdb/client/std/Compat.java|10|14|71.43%|
|:large_blue_circle:|io/questdb/client/impl/QueryClientPool.java|26|36|72.22%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/OrphanScanner.java|47|52|90.38%|
|:large_blue_circle:|io/questdb/client/impl/SenderPool.java|175|194|90.21%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpSpscQueue.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/BackgroundDrainerPool.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SlotLock.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/std/Decimal64.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/std/FdBig.java|13|13|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/std/Numbers.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|8|8|100.00%|

bluestreak01:
## Review (level 3, mission-critical pass)

Built the change-surface map with real searches, traced the cap-check invariant and every reclaim/recovery/close path, compiled on JDK 25, ran the 72 new/changed tests (green), confirmed CI green on HEAD (`d59a573`) for both the JDK 8 source-of-truth job and the JDK 25 smoke job, and **independently proved the JDK 11+ runtime `FdBig`→`jdk.internal.math.FDBigInteger` export path works** — the one runtime path this repo's CI does *not* exercise.

This is a careful, well-reasoned PR. I reproduced the central conclusions of the existing self-review independently and found **two substantive items it missed** (a hot-path allocation regression, and the exact test count), plus corroboration of its coverage gap.

### Critical

**None.** I specifically tried to break:

- **The capacity invariant.** Verified `sum(all.size + inFlightCreations + closingSlots + leakedSlots + recoveringSlots) == popcount(slotInUse)`: every `slotInUse[i]=true` write (263, 442, 1108) pairs with exactly one counter increment under `lock`, every clear (459, 489, 1122) with a decrement. So `sum < maxSize ⟺ a free index exists`, making `allocateSlotIndex()`'s throw genuinely unreachable.
- **The recovery/close races (C1)** — `markClosing()` → `housekeeper.stop()` ordering; the daemon join can't time out except on the single documented black-holed-connect residual window, and the housekeeper is a daemon so that window can't hang JVM exit.
- **`slotLockReleased` visibility** — non-volatile, but every write and read is on the same thread in all three reclaim paths.

### Moderate

**M1 — `FdBig` wrapper adds per-op allocations on the ILP double-serialization hot path** (in-diff `Numbers.java`/`FdBig.java`; realized via out-of-diff caller `AbstractLineSender`/`CharSink.put(double)`).
The reroute of `Numbers.appendDouble0`'s big-integer slow path through the new `FdBig` wrapper allocates an extra wrapper per op: `valueOfPow52`×3 + `valueOfMulPow52` + `leftShift` + **one `FdBig` per `multBy10()` loop iteration** (up to ~17) — roughly doubling the object count. I verified it is reachable per-row: `LineTcpSenderV1`/`LineHttpSenderV1`/V2 `doubleColumn` → `CharSink.put(double)` → `Numbers.append(sink, double)` → `appendDouble0` → the `FdBig` branch (taken for large-magnitude / 17-significant-digit doubles, `bBits>=64`).
- *Nuance (why Moderate):* this path **already** allocated `FDBigInteger` before the PR — never zero-GC — so it's a constant-factor worsening of an already-allocating slow fallback, not a new zero-GC violation. Most doubles take the primitive fast paths and are unaffected. But "Behaviour is unchanged" is true only for *output*, not the *allocation profile*, which changes on a per-row ingestion path.
- *Suggested fix:* drop the wrapper. Expose JDK-specific **static** bridge methods that take/return `Object` (the `FDBigInteger`) and cast internally; `Numbers` carries `Object`-typed locals and never names `FDBigInteger`. Compiles cross-JDK with **zero** added allocation versus the original.

**M2 — Production SF-recovery wiring is untested end-to-end** (corroborates the self-review's M1). `PoolHousekeeper` recovery-driving lines are 0% covered; `QuestDBImpl.close()` is 45%. The C1 test mimics the housekeeper with a hand-rolled `Thread`, not the real `PoolHousekeeper.runLoop()`. I traced the real wiring (`runLoop` → `runStartupRecoveryStep`, skip-idle-wait branch, `stop` re-check, `markClosing()`-before-`stop()`, the 2s join) and found it **correct** — so this is coverage/test-quality, not a latent bug. For a mission-critical close path it still warrants one integration test: a real handle against a silent server with seeded stranded slots, asserting `close()` returns within `STOP_TIMEOUT_MILLIS` and data is recovered against an acking server.

**M3 — Scope: the JDK 8 target-lowering is bundled with the SF fix.** The build-target change (the `Numbers`/`FdBig` reroute, `Unsafe` export removal, two Maven profiles, per-JDK source roots, new CI) is independent and has its own wide blast radius. The body itself notes M‑B recommended splitting; I reaffirm it for bisectability — not a code defect.

### Minor

- **m1 — Test-count claim off by 2.** Body says `SenderPoolSfTest (27)` / `27/27 ✅`; the suite has **29** `@Test` methods and surefire runs **29**. (The self-review's "28" is also off.)
- **m2 — No `Fixes #NNN`** despite the `bug` label.
- **m3 — New `LineSenderBuilder` introspection methods are frozen public API** (`getConfiguredSenderId`/`getConfiguredSfDir`/`isStoreAndForwardEnabled`/`orphanDrainExcludeManagedSlots`) exposing pool-internal slot-namespace concepts on a `public final` nested type. Acceptable, but now a compatibility commitment.
- **m4 — Unrelated test-client refactor bundled in** (`QwpAllocationTestClient` drops `inFlightWindow`).

### Downgraded (false positives, verified against source)

- **`FdBig` runtime `IllegalAccessError` on JDK 11+ (class-init vs. export ordering)** — FALSE. CI never exercises it (JDK 8: export is a no-op; JDK 25: `-DskipTests`). I ran the built module on JDK 25 from an unnamed module with **no** `--add-exports`: hard doubles serialize correctly via `Compat.exportFdBigInteger()`. The static block exports before the first `FDBigInteger` reference resolves.
- **`recoveringSlots` leaked if the pool closes mid-drain** — FALSE. No early return between reserve (443) and the matching decrement (458/475); `drainCandidateSlotForRecovery` catches all `Throwable`.
- **`allocateSlotIndex()` throws under concurrent borrow/recovery** — FALSE. Unreachable given `sum == popcount(slotInUse)`.
- **`discardBroken()` skips `delegate().close()` when already closed → flock leak** — FALSE. In that ordering `s` is still in `all` so `close()`'s snapshot closes it; in the other ordering `discardBroken` removed it and closes it itself. Disjoint, no double-close.
- **`borrow()` error path leaks `inFlightCreations`/slot index or double-unlocks** — FALSE. The `finally` guards on `lock.isHeldByCurrentThread()`; the catch undoes the reservation idempotently.
- **Daemon-thread join timeout hangs JVM exit** — FALSE. `questdb-pool-housekeeper` is `setDaemon(true)`.
- **JDK 8 syntax/API regressions** — whole tree clean (no `var`, `List.of`, text blocks, `String.repeat/strip/isBlank`, `ProcessHandle`, or `Thread.onSpinWait` outside the `java11` shim); the `Provider(String,double,String)` ctor swap is correctly applied. JDK 8 CI (full suite) is green.

### Summary

**Verdict: approve** — with M1 (hot-path allocation) and M2 (production-path test) worth addressing before merge; M3 (split the JDK-8 work) at the author's discretion; minors trivial.

- Original collision bug is real and the fix is correct (distinct `<base>-<index>` ids); flock-safe lifecycle, Throwable-widened teardown, and deferred/cancellable recovery are sound; the central capacity invariant holds under concurrency. 72 new/changed tests pass locally; CI green on JDK 8 + JDK 25.
- Draft findings: ~13 considered → 6 verified (3 Moderate, 4 Minor incl. corroborations), 7 dropped as false positives/unreachable after source verification.
- In-diff vs out-of-diff: 4 in-diff, 2 out-of-diff (M1's hot-path caller `AbstractLineSender`/`CharSink`; M2's untested `PoolHousekeeper`/`QuestDBImpl.close()` wiring).
- Net new signal beyond the existing self-review: the **`FdBig` double-slow-path allocation regression on the ILP hot path**, and an empirical proof that the otherwise-CI-untested JDK 11+ runtime export actually works.

<sub>🤖 Generated with review-pr (level 3)</sub>

nwoolmer (approved):
L G T M

PR #56, opened 2026-06-24
ci(build): build and release questdb-client on JDK 8

bluestreak01:
## Problem

We recently moved `java-questdb-client` to a **JDK 8 floor** — the client ships as the Java 8 `io.questdb:questdb-client` artifact, and `ci.yml` already treats JDK 8 as the source of truth (`build-jdk8`). But the **Maven Central release pipeline** (`maven_central_release.yml`) was still on newer JDKs:

| Job | Was | Role |
|-----|-----|------|
| `resolve` | JDK 11 | reads versions |
| `build-macos` | JDK 11 | macOS native lib |
| `build-linux-x86-64` / `aarch64` / `windows` | GraalVM **25** | native libs |
| `verify` | JDK 11 | `mvn verify` — **test gate** |
| `publish` | JDK 11 | `mvn deploy` — **ships to Central** |
| `open-bump-pr` | JDK 11 | snapshot bump PR |

The POMs auto-activate a profile from the running JDK. On JDK 11 the `java11+` profile (`core/pom.xml`) kicks in and **changes what gets shipped**:

- `javac source/target` → **11** → bytecode is **class-file v55**, so a JDK 8 runtime hits `UnsupportedClassVersionError`.
- source root → `src/main/java11` (module-info + `jdk.internal.math.FDBigInteger` bridge).
- deps → slf4j 2.0.17 + logback 1.5.25 (Java 11+).

So `verify` was testing — and `publish` was shipping — a **Java 11 artifact**, contradicting the "ships as a Java 8 artifact, released from JDK 8" contract.

## Change

Move the **entire** release workflow to JDK 8.

**`setup-java` jobs (5)** — `resolve`, `build-macos`, `verify`, `publish`, `open-bump-pr`: `java-version: "11"` → `"8"`. Now the `java8` profile activates, so `verify`/`publish` build and ship the real Java 8 artifact (source/target `1.8`, `src/main/java8` shim, logback 1.3.x).

**Native-lib jobs (3)** — `build-linux-x86-64`, `build-linux-aarch64`, `build-windows-x86-64`: replace `Install GraalVM JDK 25` with `Install Temurin JDK 8` via the Adoptium *latest-GA* redirect (`api.adoptium.net/v3/binary/latest/8/ga/<os>/<arch>/jdk/hotspot/normal/eclipse`).

## Why JDK 8 is correct for the native builds

The native build consumes only `$JAVA_HOME/include/jni.h` + `include/<platform>/jni_md.h` (verified in `core/CMakeLists.txt`, lines 95–128) and produces platform machine code, not Java bytecode. The JNI ABI is stable, so building against JDK 8 headers is fine — and is the most correct choice for a Java 8 floor. The Windows cross-compile's `jni_md.h` now comes from `openjdk/jdk8u` (`jdk/src/windows/javavm/export/jni_md.h`) instead of `jdk25u`; the macros/typedefs are byte-identical.

The manylinux containers (glibc 2.17 / 2.28) run Temurin 8 fine, and the install mirrors the prior `wget`-tarball pattern (no `setup-java`, which needs Node 20 unavailable in the old container).

## Verification

- YAML validated (`yaml.safe_load`).
- Confirmed the Adoptium JDK 8 download URLs (x64 + aarch64) resolve to real tarballs.
- Confirmed the `openjdk/jdk8u` Windows `jni_md.h` path resolves and contains the expected `JNICALL`/`jint`/`jlong` definitions.
- No remaining `11` / `25` / `graalvm` / `jdk25u` references in the workflow.

## Out of scope

`rebuild_native_libs.yml` also uses GraalVM JDK 25, but it's a separate utility workflow, not the release-to-Central path. Happy to align it in a follow-up if wanted.


nwoolmer (approved):
LGTM

PR #67, opened 2026-07-10
fix(qwp): keep SF slot locked until manager worker quiesces

bluestreak01:
> **Tandem PRs:** OSS server CI — questdb/questdb#7387 · Enterprise e2e — questdb/questdb-enterprise#1127

## Summary

Fix store-and-forward shutdown so a slot is never reused while a manager or I/O worker can still touch its ring, watermark, transport, or segment files. The change also makes deferred cleanup observable by sender pools, makes close-time segment cleanup crash-safe, and keeps transient startup recovery retryable for the life of the pool.

The branch additionally adds periodic store-and-forward syncing (`sf_durability=periodic` with a new `sf_sync_interval_millis` key) for bounded-loss protection against host power loss, serializes concurrent `QuestDB.close()` callers through shutdown completion, hardens the recovery-time segment sort against quadratic input, and fixes an off-by-one in `ObjList.remove(from, to)`.

This addresses the lifecycle race behind the intermittent unsafe-memory failure in [QuestDB build 249990](https://dev.azure.com/questdb/questdb/_build/results?buildId=249990) and prevents stale workers from mutating files owned by a replacement sender.

## Root cause

`SegmentManager.deregister(ring)` removed the ring from the live registry, but the worker could already have copied its `RingEntry` into a snapshot and entered `serviceRing()`. Close then unmapped the ring and watermark, unlinked files, and released the slot flock without proving that the in-flight pass had ended.

A replacement engine could acquire the same directory while the stale pass was still provisioning, trimming, or unlinking files. The outcomes included slot corruption, data loss after restart, and mmap/SIGBUS-style failures.

### Worker-quiescence barrier

- `SegmentManager.serviceRing()` claims each entry with an atomic registered/in-service state transition.
- Entries deregistered before claim are skipped.
- Deregistration of an active pass remains visible until that pass finishes.
- Shared managers support a bounded, interrupt-preserving per-ring quiescence wait.
- Owned managers use the stronger whole-worker stop/reap barrier.
- A timed-out owned close transfers its preallocated terminal cleanup to worker exit, after the final service pass.
- Ring, watermark, unlink, and flock cleanup is protected by an exactly-once CAS so retried close and worker-exit cleanup cannot race or deadlock.

### Confirmed slot release and pool recovery

- `SlotLock.release()` explicitly unlocks through platform JNI and reports whether release was confirmed.
- `closeCompleted` and `isSlotLockReleased()` become true only after confirmed unlock.
- The sender retains incomplete engines so a later worker/I/O exit remains observable.
- Deferred completion notifies `SenderPool` immediately; parked borrowers do not wait for a timeout or housekeeper tick.
- Retired in-range and startup-recovery slots are retained and returned to capacity after late release.
- Zero-timeout and expired-budget borrows perform a final retired-slot probe before failing.

### Crash-safe segment cleanup

- Close persists the final acknowledged FSN through the still-mapped watermark before closing or unlinking segments.
- Segment enumeration completes before deletion starts.
- Fully acknowledged segments are removed in generation order and deletion stops on the first failure, leaving a safely covered contiguous residue.
- The watermark is removed only after every segment is confirmed gone.
- Recovery therefore skips acknowledged residue after an unlink failure or crash instead of replaying it.

### Periodic store-and-forward syncing

- A new durability mode, `sf_durability=periodic`, checkpoints mmap-published segment data to disk in the background, extending SF protection from process restarts to host power loss with a bounded loss window. It requires `sf_dir` and WebSocket transport.
- A new ingress config key, `sf_sync_interval_millis` (builder: `storeAndForwardSyncIntervalMillis(long)`), sets the target checkpoint cadence in milliseconds. Default: `5000` in periodic mode. Accepted range: positive and at most `Long.MAX_VALUE / 1_000_000` (the nanosecond-conversion bound); values outside it throw `sf_sync_interval_millis is out of range`.
- The builder rejects the key unless `sf_durability=periodic` (`sf_sync_interval_millis requires sf_durability=periodic`) and rejects it for non-WebSocket transports.
- The manager worker msyncs each ring's live published data on the cadence; a failed checkpoint is recorded for producer visibility and retried within one second.
- Checkpoints use checked mmap and file-descriptor barriers; segment rotation waits until the predecessor segment is durable, and hot-spare installation syncs the segment header and parent directory.
- The interval is a target, not a guarantee: JVM scheduling and storage-sync latency add to the actual loss window. `request_durable_ack=on` remains the option for end-to-end server durability.
- The README documents the key and the periodic-durability contract.

### Retryable recovery and flock release

- Managed-slot startup recovery advances its cursor only after a successful attempt; transient build/connect/drain failures remain pending for a later housekeeper tick.
- Failed flock releases use one shared retry driver rather than one thread per engine.
- The driver applies exponential backoff from 100 ms to a 5 s cap, resets after progress, and is unparked when new work arrives.
- Pool probes re-arm retry scheduling after an injected driver-start failure.

### Hardened recovery segment sort

- `SegmentRing` recovery sorts enumerated segments by unsigned `baseSeq`. The previous median-of-three Lomuto quicksort was O(N²) on organ-pipe, duplicate-heavy, and med3-killer orders — reachable only through corrupted-yet-parseable or operator-copied headers, but the quadratic stall landed before validation could reject the slot.
- The sort is now an introsort: the median-of-three fast path is unchanged, each partition path carries a `2*floor(log2(N))` budget, and ranges that exhaust it fall back to in-place heapsort (still `Long.compareUnsigned`, zero allocation), capping every adversarial pattern at O(N log N) comparisons.

### Bounded transport shutdown

- Socket and WebSocket layers expose traffic shutdown separately from final close.
- Sender close first shuts down active traffic to break a worker blocked in native send/receive, then joins the worker, then performs final socket/resource close.
- POSIX and Windows native implementations provide the same shutdown contract.

### Serialized QuestDB.close()

- `QuestDBImpl.close()` set the volatile `closed` flag before running the pool teardown chain, so a second concurrent closer could observe `closed == true` and return while the first was still draining and releasing pool resources — breaking the expectation that shutdown has completed once `close()` returns.
- `close()` is now `synchronized`: the losing caller blocks until the winner finishes teardown, then enters, sees `closed`, and no-ops. The teardown steps never re-enter `close()` on another thread, so the monitor cannot deadlock.

### ObjList.remove(from, to) backing-array fix

- `ObjList.remove(int from, int to)` filled `[pos, buffer.length - 1)` with null, leaving the final backing-array slot holding a stale object reference after removal. The fill now covers the whole tail, so removed elements cannot be retained through the last slot.

### Attachment and observability guards

- A sender cannot replace or detach an already attached cursor engine.
- Test-only lifecycle hooks expose close entry, close-drain wait, worker passes, cleanup handoff, retry progress, and final release without changing production control flow.

### Review follow-ups

- `SegmentRing.findSegmentContaining0` now binary-searches the sealed list (strictly ascending and contiguous in `baseSeq`) instead of linear-scanning it, so cursor repositioning on reconnect is O(log live-sealed) even under a producer-outpaces-drain backlog. Boundary-matrix guard tests in `SegmentRingTest` were pinned green against the pre-rewrite linear implementation.
- `AckWatermark` now routes its lifetime mmap/munmap through the injected `FilesFacade`, matching `MmapSegment`, which makes watermark-mmap fault injection possible from test facades. `AckWatermarkTest` adds facade-routing and mmap-fault tests, and two `SegmentManager` tests replace reflective access with direct calls against the now-public `open(FilesFacade, String)`.

## Failure behavior

Safety takes precedence over capacity: if worker quiescence or slot release cannot be confirmed, the slot remains unavailable rather than being handed to a replacement. Normally, worker-exit cleanup, immediate pool notification, and the shared retry driver restore capacity once the blocking condition clears. A genuinely wedged worker or permanently failing OS unlock can keep the slot retired until process exit.

## Test coverage

Deterministic tests cover:

- shared- and owned-manager mid-pass close races;
- stale snapshot rejection and quiescence interrupt preservation;
- worker-exit cleanup handoff and exactly-once terminal cleanup;
- same-slot reacquisition only after safe release;
- delegated I/O-thread close and blocked native send/receive shutdown;
- confirmed unlock publication and unlock-failure retry;
- shared retry-driver backoff, progress reset, and start-failure recovery;
- close-time unlink/enumeration failure and watermark-preserved recovery;
- transient in-range and out-of-range startup-recovery retries;
- immediate wakeup of parked pool borrowers after deferred release;
- zero-timeout and final-timeout retired-slot probes;
- periodic-sync scheduling, checkpoint-barrier failures, PERIODIC rotation gating (verified mutant-killer: re-neutralizing the gate now fails), transient-failure recovery, and multi-segment stop-on-first-failure/retry passes (`SegmentManagerPeriodicSyncTest`); the manifest boundary monotonic clamp, verified mutant-killer (`SfManifestClampTest`); sender-facing durability-latch surfacing (`CursorSendEngineTest`); the close-path second bounded join (`SegmentManagerCloseRaceTest`); torn close-time directory enumeration (`CursorSendEngineClosePartialEnumerationTest`); a `MUTANT`-marker source tripwire (`SourceHygieneTest`); plus config acceptance and rejection of `sf_sync_interval_millis` (`SfFromConfigTest`, `WsSenderConfigHonoredTest`);
- adversarial recovery-sort orders (organ-pipe, all-duplicate, few-distinct, med3-killer, sign-bit keys) at N=16_384 with comparison-count bounds (`SegmentRingTest`);
- latch-controlled two-concurrent-closer serialization of `QuestDB.close()` (`QuestDBImplCloseTest`);
- final-backing-slot clearing in `ObjList.remove(from, to)` (`ObjListTest`); and
- pool capacity recovery across normal, startup, shared-manager, and real worker-wedge paths.

Enterprise PR #1127 adds real-server multi-slot crash recovery, committed-but-unacked replay with WAL/DEDUP, and close-drain failover across deterministic and seeded role-change schedules.

## Compatibility

The implementation retains the Java 8 language/API floor. Platform-specific slot unlock and traffic shutdown are implemented for POSIX and Windows and exercised by the tandem CI matrix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,8 +142,9 @@ You can also let the client flush batches for you with the `auto_flush_rows` / `
 `ws::addr=localhost:9000;auto_flush_rows=10000;auto_flush_interval=1000;`.
 
 **Confirm a batch is durably received.** Over QWP each flush returns a frame sequence number (FSN); `awaitAckedFsn`
-blocks until the server has acknowledged it. Rows are safe in the store-and-forward log even if the ack has not landed
-yet — they replay on reconnect.
+blocks until the server has acknowledged it. With `sf_dir`, rows in the store-and-forward log replay after reconnect
+or a producer-process restart. For periodic host-power-loss checkpoints, also configure
+`sf_durability=periodic;sf_sync_interval_millis=5000;`.
 
 ```java
 try (Sender sender = db.borrowSender()) {
@@ -259,7 +260,9 @@ try (QuestDB db = QuestDB.builder()
 List every cluster node in one `addr` server list; the single string configures both the ingest and query pools across
 all of them. On the query side, `target` selects the node role to route to (`any`, `primary`, or `replica`) and
 `failover=on` enables failover across the list. The ingest side reconnects across the same node list on its own — a
-store-and-forward sender keeps buffering rows through a failover window and never drops them.
+store-and-forward sender keeps buffering rows through a failover window and replays unacknowledged data. The default
+`memory` durability mode protects process restarts, not host power loss; use periodic durability for background disk
+checkpoints.
 
 ```java
 try (QuestDB db = QuestDB.connect(
@@ -439,7 +442,10 @@ Applied by the query pool to select and fail over between the nodes in the `addr
 | `client_id` |         | Opaque client identifier surfaced server-side for observability                          |
 
 The ingest side also accepts store-and-forward and reconnection tuning keys (`auto_flush_*`, `initial_connect_retry`,
-`reconnect_*`, `request_durable_ack`, `sf_*`, `max_frame_rejections`, `poison_min_escalation_window_millis`, …). See the
+`reconnect_*`, `request_durable_ack`, `sf_*`, `max_frame_rejections`, `poison_min_escalation_window_millis`, …).
+`sf_durability=periodic` checkpoints mmap-published data in the background; `sf_sync_interval_millis` defaults to `5000`
+in that mode. The interval is a target cadence: JVM scheduling and storage-sync latency add to the actual loss window.
+Use `request_durable_ack=on` when end-to-end server durability is also required. See the
 [QuestDB documentation](https://questdb.com/docs/) for the full reference.
 
 ## Requirements
--- a/core/src/main/c/share/files.c
+++ b/core/src/main/c/share/files.c
@@ -388,3 +388,15 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_msync
         (JNIEnv *e, jclass cl, jlong addr, jlong len, jboolean async) {
     return msync((void *) (uintptr_t) addr, (size_t) len, async ? MS_ASYNC : MS_SYNC);
 }
+
+/* Best-effort page pin. Callers treat a refusal (RLIMIT_MEMLOCK, missing
+ * privilege) as a soft downgrade, so no errno capture is required here. */
+JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_mlock0
+        (JNIEnv *e, jclass cl, jlong addr, jlong len) {
+    return mlock((void *) (uintptr_t) addr, (size_t) len);
+}
+
+JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_munlock0
+        (JNIEnv *e, jclass cl, jlong addr, jlong len) {
+    return munlock((void *) (uintptr_t) addr, (size_t) len);
+}
--- a/core/src/main/c/share/files.c
+++ b/core/src/main/c/share/files.c
@@ -236,6 +236,18 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_lock
     return flock((int) fd, LOCK_EX | LOCK_NB);
 }
 
+JNIEXPORT jint JNICALL Java_io_questdb_client_cutlass_qwp_client_sf_cursor_SlotLock_release0
+        (JNIEnv *e, jclass cl, jint fd) {
+    if (flock((int) fd, LOCK_UN) != 0) {
+        return -1;
+    }
+    /* Unlock success confirms that the slot is reusable. close() is one-shot:
+     * POSIX leaves descriptor state unspecified on EINTR, so retrying its
+     * numeric value could close an unrelated descriptor after reuse. */
+    (void) close((int) fd);
+    return 0;
+}
+
 JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_mkdir0
         (JNIEnv *e, jclass cl, jlong lpszPath, jint mode) {
     return mkdir((const char *) (uintptr_t) lpszPath, (mode_t) mode);
--- a/core/src/main/c/share/files.c
+++ b/core/src/main/c/share/files.c
@@ -124,6 +124,21 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_fsync
     return res;
 }
 
+JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_fsyncDir0
+        (JNIEnv *e, jclass cl, jlong lpszName) {
+    int fd;
+    RESTARTABLE(open((const char *) (uintptr_t) lpszName, O_RDONLY), fd);
+    if (fd < 0) {
+        return -1;
+    }
+    int res;
+    RESTARTABLE(fsync(fd), res);
+    int saved_errno = errno;
+    close(fd);
+    errno = saved_errno;
+    return res;
+}
+
 JNIEXPORT jboolean JNICALL Java_io_questdb_client_std_Files_truncate
         (JNIEnv *e, jclass cl, jint fd, jlong size) {
     int res;
--- a/core/src/main/c/share/files.c
+++ b/core/src/main/c/share/files.c
@@ -65,6 +65,13 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_openRW0
     return (jint) fd;
 }
 
+JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_openRWExclusive0
+        (JNIEnv *e, jclass cl, jlong lpszName) {
+    int fd;
+    RESTARTABLE(open((const char *) (uintptr_t) lpszName, O_CREAT | O_EXCL | O_RDWR, 0644), fd);
+    return (jint) fd;
+}
+
 JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_openAppend0
         (JNIEnv *e, jclass cl, jlong lpszName) {
     int fd;
--- a/core/src/main/c/share/net.c
+++ b/core/src/main/c/share/net.c
@@ -128,6 +128,11 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_send
     return com_questdb_network_Net_EOTHERDISCONNECT;
 }
 
+JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_shutdown
+        (JNIEnv *e, jclass cl, jint fd) {
+    return shutdown((int) fd, SHUT_RDWR);
+}
+
 JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_recv
         (JNIEnv *e, jclass cl, jint fd, jlong ptr, jint len) {
     ssize_t n;
--- a/core/src/main/c/share/net.c
+++ b/core/src/main/c/share/net.c
@@ -130,7 +130,8 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_send
 
 JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_shutdown
         (JNIEnv *e, jclass cl, jint fd) {
-    return shutdown((int) fd, SHUT_RDWR);
+    const int result = shutdown((int) fd, SHUT_RDWR);
+    return result == -1 && errno == ENOTCONN ? 0 : result;
 }
 
 JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_recv
--- a/core/src/main/c/share/net.h
+++ b/core/src/main/c/share/net.h
@@ -82,6 +82,14 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_send
 
 /*
  * Class:     io_questdb_client_network_Net
+ * Method:    shutdown
+ * Signature: (I)I
+ */
+JNIEXPORT jint JNICALL Java_io_questdb_client_network_Net_shutdown
+        (JNIEnv *, jclass, jint);
+
+/*
+ * Class:     io_questdb_client_network_Net
  * Method:    getSndBuf
  * Signature: (I)I
  */
--- a/core/src/main/c/windows/errno.h
+++ b/core/src/main/c/windows/errno.h
@@ -6,4 +6,10 @@ static DWORD dwTlsIndexLastError = 0;
 
 void SaveLastError();
 
+/* Returns the per-thread error code most recently stored by SaveLastError().
+ * Both functions are defined in os.c, the only translation unit whose static
+ * dwTlsIndexLastError copy is initialised by TlsAlloc() in DllMain; every
+ * other includer's copy of the variable is an unused zero. */
+DWORD GetSavedLastError();
+
 #endif //ZLIB_ERRNO_H
--- a/core/src/main/c/windows/files.c
+++ b/core/src/main/c/windows/files.c
@@ -77,11 +77,16 @@ static jint open_file(const char *utf8Path,
     }
     HANDLE h = CreateFileW(wide, desiredAccess, shareMode, NULL,
                            creationDisposition, flagsAndAttributes, NULL);
-    free(wide);
     if (h == INVALID_HANDLE_VALUE) {
+        // Save the CreateFileW failure code BEFORE free(): the thread's
+        // last-error value may be overwritten by any subsequent API/CRT call
+        // (HeapFree can SetLastError), and callers such as fsyncDir0 rely on
+        // the saved value being the CreateFileW error.
         SaveLastError();
+        free(wide);
         return -1;
     }
+    free(wide);
     return HANDLE_TO_FD(h);
 }
 
@@ -237,10 +242,12 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_fsyncDir0
         // directory handle with ERROR_ACCESS_DENIED. Directory-entry durability
         // on NTFS is provided by metadata journaling ($LogFile), so degrade to
         // best-effort success here rather than hard-failing the SF durability
-        // path. open_file() already recorded the error via SaveLastError(); a
-        // genuine failure such as a missing directory (ERROR_PATH_NOT_FOUND)
-        // still propagates as fatal.
-        if (GetLastError() == ERROR_ACCESS_DENIED) {
+        // path. Consult the error open_file() saved at the CreateFileW failure
+        // point rather than the live GetLastError(): the intervening free()
+        // and return path inside open_file() are not guaranteed to preserve
+        // the thread's last-error value. A genuine failure such as a missing
+        // directory (ERROR_PATH_NOT_FOUND) still propagates as fatal.
+        if (GetSavedLastError() == ERROR_ACCESS_DENIED) {
             return 0;
         }
         return -1;
--- a/core/src/main/c/windows/files.c
+++ b/core/src/main/c/windows/files.c
@@ -360,18 +360,24 @@ JNIEXPORT jlong JNICALL Java_io_questdb_client_std_Files_length0
                            OPEN_EXISTING,
                            FILE_ATTRIBUTE_NORMAL,
                            NULL);
-    free(wide);
     if (h == INVALID_HANDLE_VALUE) {
+        /* Save the CreateFileW failure code BEFORE free(): any subsequent
+         * API/CRT call (HeapFree can SetLastError) may overwrite the
+         * thread's last-error value -- same pattern as open_file(). */
         SaveLastError();
+        free(wide);
         return -1;
     }
+    free(wide);
     LARGE_INTEGER sz;
     BOOL ok = GetFileSizeEx(h, &sz);
-    CloseHandle(h);
     if (!ok) {
+        /* Save before CloseHandle() can clobber the last-error value. */
         SaveLastError();
+        CloseHandle(h);
         return -1;
     }
+    CloseHandle(h);
     return (jlong) sz.QuadPart;
 }
 
@@ -414,11 +420,13 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_mkdir0
         return -1;
     }
     BOOL ok = CreateDirectoryW(wide, NULL);
-    free(wide);
     if (!ok) {
+        /* Save before free() can clobber the last-error value. */
         SaveLastError();
+        free(wide);
         return -1;
     }
+    free(wide);
     return 0;
 }
 
@@ -447,11 +455,13 @@ JNIEXPORT jboolean JNICALL Java_io_questdb_client_std_Files_remove0
     } else {
         ok = DeleteFileW(wide);
     }
-    free(wide);
     if (!ok) {
+        /* Save before free() can clobber the last-error value. */
         SaveLastError();
+        free(wide);
         return JNI_FALSE;
     }
+    free(wide);
     return JNI_TRUE;
 }
 
@@ -469,12 +479,15 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_rename0
         return -1;
     }
     BOOL ok = MoveFileExW(oldW, newW, MOVEFILE_REPLACE_EXISTING);
-    free(oldW);
-    free(newW);
     if (!ok) {
+        /* Save before free() can clobber the last-error value. */
         SaveLastError();
+        free(oldW);
+        free(newW);
         return -1;
     }
+    free(oldW);
+    free(newW);
     return 0;
 }
 
@@ -512,11 +525,13 @@ JNIEXPORT jlong JNICALL Java_io_questdb_client_std_Files_findFirst0
     pattern[pathLen] = '\0';
 
     wchar_t *wide = utf8_to_wide(pattern);
-    free(pattern);
     if (!wide) {
+        /* Save before free() can clobber the last-error value. */
         SaveLastError();
+        free(pattern);
         return 0;
     }
+    free(pattern);
 
     qdb_find_t *find = (qdb_find_t *) malloc(sizeof(qdb_find_t));
     if (!find) {
@@ -524,12 +539,14 @@ JNIEXPORT jlong JNICALL Java_io_questdb_client_std_Files_findFirst0
         return 0;
     }
     find->handle = FindFirstFileW(wide, &find->data);
-    free(wide);
     if (find->handle == INVALID_HANDLE_VALUE) {
+        /* Save before free() can clobber the last-error value. */
         SaveLastError();
+        free(wide);
         free(find);
         return 0;
     }
+    free(wide);
     find->hasEntry = 1;
     win_findname_to_utf8(find);
     return (jlong) (uintptr_t) find;
--- a/core/src/main/c/windows/files.c
+++ b/core/src/main/c/windows/files.c
@@ -215,6 +215,25 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_fsync
     return 0;
 }
 
+JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_fsyncDir0
+        (JNIEnv *e, jclass cl, jlong lpszName) {
+    jint fd = open_file((const char *) (uintptr_t) lpszName,
+                        GENERIC_READ | GENERIC_WRITE,
+                        FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+                        OPEN_EXISTING,
+                        FILE_FLAG_BACKUP_SEMANTICS);
+    if (fd < 0) {
+        return -1;
+    }
+    if (!FlushFileBuffers(FD_TO_HANDLE(fd))) {
+        SaveLastError();
+        CloseHandle(FD_TO_HANDLE(fd));
+        return -1;
+    }
+    CloseHandle(FD_TO_HANDLE(fd));
+    return 0;
+}
+
 JNIEXPORT jboolean JNICALL Java_io_questdb_client_std_Files_truncate
         (JNIEnv *e, jclass cl, jint fd, jlong size) {
     FILE_END_OF_FILE_INFO eof;
--- a/core/src/main/c/windows/files.c
+++ b/core/src/main/c/windows/files.c
@@ -103,6 +103,15 @@ JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_openRW0
                      FILE_ATTRIBUTE_NORMAL);
 }
 
+JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_openRWExclusive0
+        (JNIEnv *e, jclass cl, jlong lpszName) {
+    return open_file((const char *) (uintptr_t) lpszName,
+                     GENERIC_READ | GENERIC_WRITE,
+                     FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+                     CREATE_NEW,
+                     FILE_ATTRIBUTE_NORMAL);
+}
+
 JNIEXPORT jint JNICALL Java_io_questdb_client_std_Files_openAppend0
         (JNIEnv *e, jclass cl, jlong lpszName) {
     jint fd = open_file((const char *) (uintptr_t) lpszName,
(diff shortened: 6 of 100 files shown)


bluestreak01:
> **Scope note:** This is the **tandem PR review** of Enterprise `questdb-enterprise#1127` + OSS `questdb/questdb#7387`, posted here because this client PR (`java-questdb-client#67`) is the third leg of the tandem and its revision (`9c9e1ddbda9a`) is what those two PRs pin. The client change was **verified for store-and-forward-contract compliance** in that context (result below) but was **not** given a full standalone line-by-line review — that belongs to this PR's own review. The findings below are almost entirely on the ENT/OSS side; the client-relevant result is the SF-contract clearance in the header and the "Downgraded" section.

**Client SF-contract clearance:** the bumped revision keeps the steady-state SF drainer reconnect **unbounded** — `reconnectMaxDurationMillis` is explicitly *"NOT consulted by the background loop: Invariant B removed the wall-clock give-up from connectLoop"* and only bounds the non-lazy initial connect (which the SF contract permits); per-attempt backoff is capped while the retry loop stays unbounded. No reconnect time budget, no hard-fail on transient outage, and no watermark-advance-past-NACK was observed in the drainer path. Compliant.

---

## Review — QWP multi-slot recovery & close-drain failover (tandem)

**Tandem review** of Enterprise **#1127** + OSS **questdb/questdb#7387** (client **questdb/java-questdb-client#67** SF-contract-checked). Full mission-critical pass: 14 review dimensions across independent fresh-context reviewers (Rust N/A — no `.rs`); every load-bearing finding re-verified against source.

### What this PR is
An **e2e test PR** (ENT, ~3.5k lines of harness/tests) plus a submodule bump carrying **one OSS server behavioral change**: `gracefulCloseAndDisconnect` → **`gracefulCloseAndDrain`**, a bounded post-CLOSE read-drain so a server-initiated fatal WebSocket CLOSE lingers (FIN + keep reading/discarding) instead of closing the fd under in-flight client bytes. An fd close under in-flight bytes forces a TCP RST that destroys the final durable ACK in the peer's recv queue → SF client replays committed-but-unacked batches → duplicates on non-`DEDUP` tables.

**Tandem chain verified consistent** (despite stale descriptions — see Minor): OSS #7387 head `1898c786a6` == ENT-pinned `questdb@1898c786a6` → `java-questdb-client@9c9e1ddbda9a` == OSS transitive bump. **Client SF-contract sanity:** the steady-state drainer reconnect is **unbounded** (`reconnectMaxDurationMillis` is "NOT consulted by the background loop"; it only bounds the non-lazy initial connect — permitted); per-attempt backoff capped, retry loop unbounded — compliant with the store-and-forward invariants.

---

### Critical
**None.** The production change was deep-traced end-to-end and confirmed correct; its fix ships with an effective regression test. No confirmed correctness, data-integrity, resource, concurrency, or performance/IO defect on a reachable production path remains open. (One "Blocker" candidate was raised and dropped after verification — see **Downgraded**.)

---

### Moderate
Test-efficacy / coverage-clarity items (they do not touch production correctness/perf). All are cheap to fix in this PR.

**M1 — [ENT] `test_close_drain_failover.py` is not a regression guard for the RST-destroys-ACK bug (misattributed coverage).** The test withholds the 40-row backlog from node A via paused forward gates before demote (its own docstring: "forward gates retain the backlog before both endpoints, then A demotes while those bytes cannot reach it") and durably-acks the 30-row baseline beforehand. So **A holds no committed-but-unacked data at demote** — there is no final durable ACK for an RST to destroy. Reverting the production hunk (immediate RST) yields the identical dense `[0,70)` on B (client reconnects, replays from `ackedFsn+1`); the `close_pending`/ordering assertions are also insensitive to RST-vs-clean-CLOSE. It is a valid, well-built test of the *close-rides-failover choreography* and should be kept. Fix: either (a) add a deterministic variant where the backlog is durably **committed on A before demote** with the final durable-ACK withheld by a paused reverse gate — a reverted RST then replays committed rows and breaks the dense oracle on the non-DEDUP table; or (b) adjust the docstring so it isn't represented as the RST-fix witness and cite the actual guards (OSS `QwpServerCloseDrainTest` + pre-existing ENT `SqlFailoverQwpDeferredCloseExactlyOnceTest`). *The RST fix itself IS covered (see Coverage map) — this is an attribution gap, not an uncovered bug.*

**M2 — [OSS] Peer-close fast-exit (`recv < 0`) is executed but never asserted.** In `QwpIngressUpgradeProcessor.resumeRecv` drain branch, the `drained < 0 → ServerDisconnectException` path (the *common* conformant-client exit) is only implicitly exercised at `QwpServerCloseDrainTest.testFatalCloseLingersWhileStreamingPeerDrainsItsReceiveQueue` socket-close teardown; nothing asserts the server disconnects on the peer FIN. A regression (fall-through to would-block/park) would surface only non-deterministically via `assertMemoryLeak`/pool halt. Fix: after closing the client socket, assert the server tears the connection down promptly (poll connection count → 0, or that a fresh connect succeeds, within a bounded window).

**M3 — [OSS] The `SEND_STATE_RESUME_CLOSE` send-drain path is UNTESTED.** Arming the drain from a *partially-flushed* CLOSE frame (`resumeSend` RESUME_CLOSE → `gracefulCloseAndDrain`) needs a PeerIsSlow under a small send-fragmentation cap; `QwpServerCloseDrainTest` uses default large send buffers, so the CLOSE always leaves via the READY path. This is also the only path where the `drainBufferedFrames` `!isCloseDraining()` guard matters. Fix: add an e2e case forcing `getForceSendFragmentationChunkSize()` below the CLOSE-frame size (as `QwpUpgradeRejectFragmentationTest` already does for handshake rejects) to exercise resume-close arming + buffered-frame discard.

---

### Minor
- **[OSS] `resumeRecv` drain loop does not re-check the deadline in-loop** (optional hardening). The `while ((drained = socket.recv(...)) > 0)` discards to would-block and checks `isCloseDrainExpired()` only on re-entry. This is the **same loop shape as the pre-existing `HttpConnectionContext.drainRecvBuffer()`** and the new code yields to the dispatcher between read events, so it is self-limiting — but a per-N-iteration `isCloseDrainExpired()` recheck + bounded chunk would make the 5s budget a hard guarantee even against a continuous within-invocation flood. Not a regression.
- **[OSS] Stale javadoc on `sendDeferredFatalClose`:** still says "half-closes the write side and raises ServerDisconnect" — it now calls `gracefulCloseAndDrain` and returns into the read-drain. Update the one-liner.
- **[ENT] `QwpSidecarMain.connectPool` cleanup NPEs on null `senders`:** the `catch` does `for (Sender s : senders)` before `db.close()`; if `new Sender[count]` itself throws (pathological `CONNECT_POOL <n>` / OOM), `senders` is null → NPE masks the real error and leaks the connected `QuestDB` facade, defeating the method's own comment. Test-only + unreachable in practice (`count` validated `>=1`, test-controlled). Guard with `if (senders != null) { … }` so `db.close()` always runs.
- **[OSS] `closeDrainDeadline` defensive reset:** reset only in `onDisconnected()` (safe under current teardown wiring; mirrors the shipped `roleChangeCloseDeferredDeadline` pattern). A defensive reset in `of()` would harden against future teardown-path changes that could otherwise start a reused pooled context with `isCloseDraining()==true`.
- **[BOTH] Stale submodule commit hashes in both PR descriptions** — ENT body cites `questdb→d078ef13`/`client→a167e650`; OSS body cites `ENT pins cfece81562`/`client→fee9903ca6fc`; the actual final pins are `1898c786a6`/`9c9e1ddbda9a`. Descriptions-only; refresh before merge so provenance verification isn't misled.
- **[ENT] Scope creep:** the PR adds `.pi/skills/fix-pr/SKILL.md` (430 lines of agent tooling) + `.gitignore` `.pi-subagents/`, unrelated to the QWP change. Flag for author decision (separate tooling PR / personal config) unless the team has decided to vendor `.pi/skills/`.
- **[OSS] Duplicated test helpers:** `QwpServerCloseDrainTest.createMaskedFrame`/`performWebSocketHandshake` are byte-for-byte copies of the same helpers in `QwpWebSocketProtocolTest` (both extend `AbstractQwpWebSocketTest`). Hoist into the shared base to avoid framing-fix drift.
- **[ENT] `ReverseTrafficGate._pause()` has no timeout** — waits unbounded for an in-flight `sendall` on a backpressured peer; a misuse surfaces as a whole-suite hang rather than a clean assertion. **[ENT] `QwpSidecarMain.Client` methods not alphabetical within kind** (`close`/`closeWithWitness`/`closeQuietly`) — auto-sort reconciles.

---

### Downgraded (false positives / refuted after source verification)
- **"Blocker: post-CLOSE drain loop pins an IO worker indefinitely / DoS / regression."** *Refuted.* (1) The unbounded `while (socket.recv(recvBuffer, recvBufferSize) > 0)` is **not new** — the pre-PR `gracefulCloseAndDisconnect` ran the identical loop via `context.drainRecvBuffer()`, so no regression in loop shape. (2) The new code **yields between read events**: on would-block it `return`s → `handleProtocolSwitchedRecv` throws `registerDispatcherRead()`, so the worker services other connections between drain reads — it does not monopolize across the 5s. (3) Non-blocking `recv` reaches would-block far faster than any real network delivers (drain = memcpy ≫ link rate); the legit demote-time streamer stops and closes within ms of observing the CLOSE (→ `recv<0` exit). The "re-check deadline in-loop" idea survives as a Minor hardening, not a blocker.
- **"CLOSE/CLOSE_BEGIN on null client now replies ERR (was OK)"** — intentional test-harness behavior change, no production impact.
- **OSS "formatting" files carry hidden behavior** — refuted: `ColumnType`, `Micros`/`Dates`/`Nanos`, `CharSink`, `FunctionParser`, `SqlCodeGenerator`, `Concurrent*HashMap`, etc. (16 files) are **pure nested-ternary reindentation** — every changed line differs from its pair only in leading whitespace. EXEMPT.
- **Client bump imposes a reconnect budget on the SF drainer** — refuted: `reconnectMaxDurationMillis` is explicitly "NOT consulted by the background loop" (bounds only the non-lazy initial connect, which the SF contract permits).

---

### Coverage map (OSS QWP close-drain — the only production behavioral change)
| # | Branch / path | Test | Fails on regression? | Verdict |
|---|---|---|---|---|
| 1 | drain deadline expiry → disconnect | e2e `testCloseDrainDeadlineBoundsLingerAgainstLiveWriter` (`cutOffMillis>=3000 && <=20000`) + unit `testCloseDrainLifecycle` | Yes — revert (immediate RST) → cutoff ≈0–300ms fails `>=3000` | **TESTED** |
| 2 | `recv>0` discard loop (linger, no RST) | e2e `testFatalCloseLingersWhileStreamingPeerDrainsItsReceiveQueue` | Yes — revert → RST → follow-up write throws | **TESTED** |
| 3 | `recv==0` park → re-register read | e2e tests 1 & 2 (idle gaps) | implicit only | weak |
| 4 | `recv<0` peer-close → disconnect | e2e test 2 teardown | **not asserted** | **UNTESTED (M2)** |
| 5 | send-path RESUME_CLOSE → drain arm | — | — | **UNTESTED (M3)** |
| 6 | `sendFatalClose` arming | e2e tests 1 & 2 (TEXT reject → CLOSE 1003) | Yes | **TESTED** |
| 7 | deferred/durable-ACK-then-close arming (demote) | pre-existing ENT `SqlFailoverQwpDeferredCloseExactlyOnceTest` | Yes | **TESTED** (out-of-diff) |
| 8 | `drainBufferedFrames` skip guard | — | implicit | untested directly |
| 9 | `processWebSocketFrames` discard-rest | — | implicit | untested directly |
| 10 | `beginCloseDrain` idempotency | unit `testCloseDrainLifecycle` (deadline anchored at origin → expires at `now==TIMEOUT`) | Yes | **TESTED** |
| 11 | `onDisconnected` reset | unit `testCloseDrainLifecycle` | Yes | **TESTED** |
| 12 | `closeDrainDeadline` survives `clear()`/`clearMessageState()` | unit `testCloseDrainLifecycle` | Yes | **TESTED** |
| — | 16 OSS non-QWP files | — | pure reindentation, verified no behavior | **EXEMPT** |

---

### Summary
**Verdict — ENT #1127: approve · OSS #7387: approve**, with the Moderate coverage clarifications (M1–M3) and Minors recommended for THIS PR (all cheap; not deferred).

- **Correctness & performance gate: PASS.** The OSS close-drain change was traced end-to-end (recv semantics; recv- and send-path re-registration; state survival across `reset()`/`clear()`; pooled-reuse reset via `onConnectionClosed`→`onDisconnected`; idempotent arming; all fatal-close paths arm the drain incl. the demote durable-ACK path; FIN-after-CLOSE ordering; ~zero added steady-state cost). No open production defect. The single "Blocker" was refuted (pre-existing loop shape + yields between events + self-limiting).
- **Test gate: PASS.** The fix ships with a regression test whose failure link is verified by revert-reasoning: OSS `QwpServerCloseDrainTest` fails pre-fix (immediate RST → cutoff `<3000` / IOException). The state machine has a strong non-vacuous unit test. The ENT e2e suite complies with the SF store-and-forward test-application checklist (no producer-visible role errors as write-gate evidence; replay-aware oracles via withhold-from-A construction and a `DEDUP` table; acknowledged-boundary ordering, not timing).
- **Findings:** ~14 draft findings; **1 major false-positive dropped** (drain-loop Blocker) + 3 other refutations. Split: **0 Critical · 3 Moderate · ~8 Minor**. Essentially all **in-diff**; the drain-loop shape is **pre-existing** (`drainRecvBuffer`) and the demote regression witness (row 7) is a pre-existing out-of-diff test. Cross-context (egress processor, framework re-registration, pooled-context reuse, client SF drainer) all cleared.
- **Highest-value residual:** M1 — attribute the RST-fix regression coverage correctly (OSS `QwpServerCloseDrainTest` + pre-existing `SqlFailoverQwpDeferredCloseExactlyOnceTest`), and consider the committed-on-A-before-demote variant so the ENT deterministic test also discriminates the fix.

<sub>🤖 Generated with tandem PR review (level 3).</sub>

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 2298 / 2607 (88.15%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/MmapSegmentCorruptionException.java|2|4|50.00%|
|:large_blue_circle:|io/questdb/client/std/Files.java|14|22|63.64%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|121|155|78.06%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|279|330|84.55%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|60|70|85.71%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SfManifest.java|106|122|86.89%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentRing.java|518|592|87.50%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentManager.java|356|406|87.68%|
|:large_blue_circle:|io/questdb/client/Sender.java|27|31|87.10%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SlotLock.java|65|72|90.28%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/BackgroundDrainer.java|31|34|91.18%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/MmapSegment.java|186|204|91.18%|
|:large_blue_circle:|io/questdb/client/impl/SenderPool.java|346|373|92.76%|
|:large_blue_circle:|io/questdb/client/network/JavaTlsClientSocket.java|20|21|95.24%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/AckWatermark.java|76|80|95.00%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SfSanitizedResidueException.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/network/NetworkFacade.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SfRecoveryException.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/network/NetworkFacadeImpl.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/std/ObjList.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/std/DefaultFilesFacade.java|7|7|100.00%|
|:large_blue_circle:|io/questdb/client/std/FilesFacade.java|9|9|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SfOperationalException.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SenderErrorDispatcher.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/network/PlainSocket.java|5|5|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SlotLockContentionException.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/impl/SenderSlot.java|4|4|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SenderConnectionDispatcher.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/impl/QueryClientPool.java|35|35|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SenderProgressDispatcher.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpQueryClient.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/client/network/Socket.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/impl/QuestDBImpl.java|2|2|100.00%|

bluestreak01:
## Review — `fix(qwp): keep SF slot locked until manager worker quiesces` (level 3)

head `ab5a13a0` · base `5cc68a91` · 100 files, +23,745/−2,050

Reviewed **only this client PR** (per request). The tandem PRs — OSS server-CI questdb/questdb#7387 and enterprise e2e `questdb-enterprise#1127` — were **not** reviewed here. Run at level 3: 7 parallel fresh-context reviewers partitioned across the diff (segments/quiescence · pool/SF-drainer · transport/close · periodic+native · performance · fresh-adversarial · tests), plus per-finding source verification.

### Verdict: approve

No Critical, no blocker, no confirmed data-loss/correctness/resource/concurrency/performance defect on any reachable production path. Unusually defensive, self-documenting code with strong crash-safety reasoning and a would-fail-if-reverted regression test behind every key behavior. Items below are Minor/optional polish and platform residual-risk notes; none gates the merge.

### Verified correct (highlights)
- **Worker-quiescence race fix** — the `RingEntry` `REGISTERED→IN_SERVICE→DEREGISTERED_IN_SERVICE→DEREGISTERED` CAS is race-free; `serviceRing()` publishes `inService` before the claim CAS, a deregistered stale snapshot loses the CAS and touches nothing, and there is no lost wakeup on the quiescence wait. A stale worker cannot touch ring/watermark/file after `awaitRingQuiescence` returns true.
- **Crash-safe cleanup ordering** — final acked FSN is persisted through the still-mapped watermark (`write`+`sync`+`fsyncDir`) *before* `ring.close()`/any unlink; segments are enumerated fully, unlinked in generation order, stop-on-first-failure, and the watermark file is removed *only* after the post-unlink `fsyncDir` confirms every segment gone. A crash mid-unlink leaves the durable watermark covering residual segments → recovery skips them, no acked replay, no demand for an unlinked segment.
- **SF drainer invariants** — no transport/connect/role error escapes a running drainer onto a producer or `borrow()`; primary reconnect is unbounded (`while (!stopRequested)`, exponential backoff with a per-attempt cap only); `.failed` quarantine fires only on the sanctioned terminals (auth, non-421 upgrade reject, capability-gap settle-budget exhaustion, proven durable corruption); no transient class (role reject / transport error) burns the settle budget.
- **Pool recovery** — an index returns to the free set only after a *confirmed* flock release; no capacity double-count or permanent leak; rearm handoff free of lost-wakeup/double-notify.
- **Transport shutdown / close** — `Net.shutdown` sound on POSIX (`SHUT_RDWR`, `ENOTCONN→0`) and Windows (`CancelIoEx`+`SD_BOTH`); the 30 s close backstop can neither hang forever nor proceed before the worker is joined (no reintroduced race); `synchronized QuestDB.close()` has no self-deadlock (reentrant monitor, teardown never re-enters cross-thread).
- **SfManifest** — double-slot alternation + CRC32c + monotonic clamp is tear-tolerant; fd is closed exactly once across create/open/quarantine-failure (no double-close, no leak). Native: `O_EXCL`/`CREATE_NEW` atomic, errno saved at the failure point, no fd/handle leak.
- **Performance** — all four claims verified real, correct, and optimal: `SegmentRing` recovery sort is a genuine `O(N log N)` introsort (median-of-three + `2·⌊log₂N⌋` budget + zero-alloc heapsort fallback, `Long.compareUnsigned` throughout); `findSegmentContaining0` binary search is `O(log n)` with no off-by-one; periodic-sync sealed-prefix skip collapses steady state to `O(1)`; `tryAppend` fuses the per-frame CRC into one contiguous native call (byte-identical result).

### Minor
1. **[pre-existing] `Sender.build()` surfaces an unchecked non-`LineSenderException` on a build-time filesystem fault.** `Sender.java:1558` `new CursorSendEngine(...)` sits outside any `LineSenderException` translation (the `try` starts only at the following `QwpWebSocketSender.connect`); the ctor now throws `SfOperationalException` (`CursorSendEngine.java:386,440`) when `AckWatermark.open` returns null (the intended fail-closed hardening). Verified pre-existing: the base-commit `build()` had the same unwrapped construction and the base ctor already threw `MmapSegmentException` (a `RuntimeException`) from construction, so `build()` callers were already exposed to unchecked non-`LineSenderException` types — `SfOperationalException`'s javadoc even states it "preserves the existing caller-visible construction-failure contract." Behaviorally inert w.r.t. data/resources/perf (the build fails either way), no caller-visible regression vs. master. Optional: wrap the construction in `catch (RuntimeException) → LineSenderException` to match the adjacent SF-setup throws.
2. **[in-diff, non-production] Direct-pool retired-slot fast-path miss.** In `SenderPool.recoverOneSlotStep`'s in-range retire branch a worker exit firing `recoverReleasedSlot` before `addRetiredSlot(...)` sees `retiredIndex == -1` and no-ops (`SenderPool.java:2119-2126`), falling back to `reprobeRetiredSlots()`. Confined to non-production: every production pool is built with `deferStartupRecovery=true` (`QuestDBImpl.java:131`), whose housekeeper reprobes every tick → self-heals; only test-only/external **direct** pools can transiently hold the retired index. No data loss. Optional: direct recovery loop `reprobeRetiredSlots()` once per cycle and correct the `recoverReleasedSlot` comment.
3. **[test quality] `SegmentManagerUnlinkFailureTest.readTotalBytes` reintroduces reflection** (`getDeclaredField("totalBytes")/"lock"`) even though this PR added `getTotalBytesForTesting()` and migrated `SegmentManagerTotalBytesRaceTest`/`TrimDeregisterRaceTest` to it. Call the seam — contradicts the PR's own "replaced reflection with typed seams" claim.
4. **[doc] `CursorSendEngine.finishClose(...)` javadoc overstates the precondition** ("caller must hold the engine monitor"); the shared flock-release retry driver correctly invokes it off-monitor and the real exclusion is the `terminalCleanupClaimed` CAS. Tighten so a future maintainer doesn't add monitor-dependent state.
5. **[cosmetic] Config transport-mismatch message wording differs** between the `storeAndForwardSyncIntervalMillis` setter ("store_and_forward is only supported for WebSocket transport") and the config-string path ("sf_sync_interval_millis is only supported…").
6. **[convention] `CursorSendEngine.unlinkAllSegmentFiles` uses `java.util.ArrayList` + `names.sort(lambda)`** instead of the project's zero-alloc `ObjList` sort. Cold close/drain path, non-capturing lambda → no data-path GC concern; convention divergence only.
7. **[test hygiene] `SegmentManagerPassBarrierBenchmark`** is a 131-line `public static void main` in `src/test` with no `@Test` (never runs under surefire, guards nothing) — fine as an opt-in benchmark, noted so it isn't mistaken for coverage. Also the `static long sortComparisons`/`nextSealedComparisons` test counters are written on the production recovery/IO path (non-volatile → racey under concurrent multi-slot recovery); test-only inaccuracy, negligible.

### Platform residual-risk notes (not diff defects — worth surfacing)
- **macOS periodic-durability is weaker than Linux/Windows.** POSIX `fsyncDir0` is `open(O_RDONLY)+fsync`, which is not a guaranteed directory-durability barrier on macOS/BSD (needs `F_FULLFSYNC`, invalid on dir fds), so `sf_durability=periodic`'s directory-entry crash-consistency (slot creation, unlink barriers) degrades on macOS. Linux/Windows unaffected.
- **POSIX `shutdown()` waking a blocking `connect()` is platform-dependent** (Linux may not wake a `SYN_SENT` connect from another thread); connect-wakeup is only fake-transport tested. Correctness preserved by the 30 s close backstop + delegated cleanup — worst case a close concurrent with a black-holed reconnect stalls up to 30 s and capacity recovers after OS SYN-retry. Latency/capacity cost, not data loss.

### Design note (confirmed intended — not a bug)
- In `sf_durability=periodic`, a **local-disk** checkpoint failure latches `durabilityFailure` and surfaces to the producer via `checkDurability()` on the next `appendOrFsn()`/`flush()`. This is a local durability fault, not a transport/server error, so it does not violate the SF "never hard-fail the producer on a transient outage" invariant (which governs transport/server errors); it is retryable and self-clears via `clearDurabilityFailure()`. Please just confirm "periodic-mode local disk hiccup throws to `Sender.flush()` and clears on retry" is the intended contract.

### Test coverage
Every behavioral change maps to a named test whose assertion fails if the production change is reverted: worker-quiescence claim gate (`SegmentManagerCloseRaceTest`), crash-safe cleanup ordering (`SegmentManagerUnlinkFailureTest` / `…CrashConsistencyTest` / `CursorSendEngineCrashConsistencyTest` / `SegmentRecoveryIntegrityTest`), confirmed slot release (`SlotLockTest`), deferred pool recovery (`SenderPoolSfFuzzTest` / `SenderPoolSfTest`), periodic sync (`SegmentManagerPeriodicSyncTest`), serialized `QuestDB.close()` (`QuestDBImplCloseTest`), `ObjList.remove(from,to)` last-slot clear (`ObjListTest` — independently re-verified), introsort + binary-search membership (`SegmentRingTest` / `SegmentRingMembershipTest`), SfManifest clamp + durability (`SfManifestClampTest` / `SegmentManagerManifestFsyncTest`). Native allocations wrapped in `assertMemoryLeak(...)`; concurrency harnesses capture-on-worker/rethrow-on-main (no swallowed `AssertionError`); error/NULL/fault paths covered. Missing coverage: none identified. Totals: **9 key behavioral changes, 9 tested, 0 UNTESTED.**

### Summary
- Correctness & performance gate: **passes** — no confirmed correctness/data/resource/concurrency/perf finding on a reachable production path.
- Test gate: **passes** — a fix PR with a verified would-fail-if-reverted regression test for every behavioral change.
- Findings: 7 verified (1 pre-existing, 6 in-diff), all Minor/optional; 0 Critical, 0 Moderate; plus 2 platform residual-risk notes and 1 confirmed-intended design note. Split: 6 in-diff / 0 out-of-diff / 1 pre-existing.

<sub>Automated level-3 review (7 parallel reviewers + source verification). Scope: java-questdb-client PR #67 only.</sub>

PR #68, opened 2026-07-15
chore(qwp): de-flake the SF recovery header-fault test

glasstiger:
**Tandem:** questdb/questdb#7404 bumps this branch into OSS core — merge together.

## What

The SF store-and-forward recovery header-fault regression test
(`MmapSegmentRecoveryFaultTest.testHeaderFaultOnMapPastEofIsSkippableAnyFilesystem`)
flaked the JDK 8 CI with:

```
java.lang.InternalError: a fault occurred in a recent unsafe memory
access operation in compiled Java code
```

The test called `MmapSegment.openExisting(ff, path)` directly from the test
lambda. Under C2, HotSpot inlines `openExisting` into the handler-less test
frame and delivers the async unsafe-access `InternalError` past
`openExisting`'s own `catch (Throwable)`, so the fault-to-`MmapSegmentException`
conversion never runs and a raw `InternalError` escapes. This is a property of
how HotSpot delivers the async "recent unsafe memory access" fault for
JIT-compiled code, not a defect in the recovery code itself: production
recovery runs once at startup (interpreted/C1), where the fault is delivered
precisely and caught.

## Change

- Route the two header-skippable tests through `SegmentRing.openExisting(...)`,
  production's real recovery caller. The faulting header read then sits behind a
  genuine, non-inlined call frame whose `catch (Throwable)` converts the fault
  to a `MmapSegmentException`, which `SegmentRing`'s per-file catch skips — so
  recovery of the sole unreadable segment returns no ring, exactly as production
  behaves.
- Add a facade-aware `SegmentRing.openExisting(FilesFacade, String, long)`
  overload so the length-injecting test facade reaches
  `MmapSegment.openExisting`. It mirrors `MmapSegment.openExisting`'s existing
  facade seam; the `FilesFacade.INSTANCE` overload production uses is unchanged.

Fail-on-revert is preserved: revert the header-block conversion and the raw
`InternalError` escapes `SegmentRing`, so the test errors instead of returning
`null`.

## Verification

The client's native lib was built locally so the tests run. Under full-suite
JIT warmup the routed header tests are stable:

- JDK 11 (pre-21, the closest available proxy for the JDK 8 CI): 6/6 runs green
- JDK 17: 13/13 runs green

JDK 8 was not available locally. JDK 11 shares the same pre-21 async-fault
delivery family as JDK 8, so it is the best available proxy; the JDK 8 CI is the
final confirmation.

## Known limitation, left unchanged

`testScanFaultOnMapPastEofIsHandledAnyFilesystem` is a separate, pre-existing
flake of the same class and is not touched by this PR. It is green on the JDK 8
CI (both `main` and the failing branch reported only the header test failing)
and passes in isolation, but fails in-suite on JDK 11/17 because the sibling
methods warm the JIT and the scan-path async fault then escapes. I verified that
neither routing through `SegmentRing` nor `-XX:CompileCommand=dontinline` fixes
it — the fault escapes every Java-level catch under C2 regardless. Stabilizing
it on JDK 11/17 would require either relaxing it to accept the raw
`InternalError` (which weakens its fail-on-revert guard) or isolating it in a
dedicated fork (fragile under Surefire fork reuse). Because it is green on the
JDK 8 CI and production is unaffected, it is left as-is here; it can be a
follow-up if JDK 11/17 local-dev stability is wanted.

## Test plan

- `mvn -pl core test -Dtest=MmapSegmentRecoveryFaultTest` — the two
  header-skippable tests exercise the real `SegmentRing` recovery path
- Verified locally under full-suite JIT warmup: JDK 11 6/6, JDK 17 13/13
- CI on JDK 8 confirms the header flake is gone on the shipping target

### DIFF
--- a/core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentRing.java
+++ b/core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentRing.java
@@ -25,6 +25,7 @@
 package io.questdb.client.cutlass.qwp.client.sf.cursor;
 
 import io.questdb.client.std.Files;
+import io.questdb.client.std.FilesFacade;
 import io.questdb.client.std.ObjList;
 import io.questdb.client.std.QuietCloseable;
 import org.jetbrains.annotations.TestOnly;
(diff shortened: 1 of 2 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 2 / 2 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/SegmentRing.java|2|2|100.00%|

PR #73, opened 2026-07-27
fix(qwp): fix store-and-forward recovery data loss and unbounded client shutdown

bluestreak01:
> [!IMPORTANT]
> **Tandem PR — merge all three together.**
> - Server: questdb/questdb#7366
> - Enterprise: questdb/questdb-enterprise#1113
> - Client: this PR
>
> questdb/questdb#7366 pins this branch through its `java-questdb-client` submodule, and questdb/questdb-enterprise#1113 pins that branch in turn. All three use the branch name `fix/qwp-deferred-close-echo-wait`.

## Problem

The QWP role-change handoff (questdb/questdb#7366) flushes a final durable ACK, sends a WebSocket CLOSE, and drains the tail before tearing down TCP so a store-and-forward producer cannot replay rows the promoted replica already owns. Reviewing that work surfaced a set of independent store-and-forward defects on the client side — SIGBUS on recovered sparse segments, silent data loss on unreadable segments, an unbounded shutdown daemon, and a slot released before its worker quiesced. This PR carries those fixes plus the test-stability work the same review required.

### Store-and-forward recovery correctness

- **Fix SIGBUS risk when appending into recovered sparse SF segments.** A segment recovered with holes could be appended into beyond its materialised extent, faulting the writer.
- **Fail closed on unreadable SF segments and enumeration errors during recovery.** Previously a segment that could not be read was skipped, silently dropping buffered rows; recovery now surfaces the failure instead of losing data quietly.
- **Avoid a double directory scan when registering a recovered SF ring.**
- **Make bulk SF trim O(S)** via a single range removal in `drainTrimmable`, instead of per-entry removal.

### Shutdown and lifecycle

- **Make engine close a worker-quiescence barrier before releasing slot resources** — the slot stayed locked until the manager worker actually stopped, closing a window where resources were freed under a live worker.
- **Bound the deferred-close daemon in `SenderPool` / `QueryClientPool`** and abandon permanently stuck engines rather than hanging shutdown indefinitely.
- Cover deferred-close caller fallbacks.

### Tests

- Deflake the close-lifecycle interrupt tests on saturated CI agents; keep the interrupted-close regression revert-proof and cover the mid-join interrupt.

### Close-code handling — added then reverted

This branch briefly classified a private-use `ROLE_CHANGE` (4001) close as orderly, to match a server that emitted 4001 on the role-change CLOSE. **Both halves are now reverted**: questdb/questdb#7366 keeps `NORMAL_CLOSURE` on the wire, and this PR strips the client-side classification.

The reasoning, recorded in `design/qwp-nack-policy-v2.md`: 4001 only let the server distinguish a genuine CLOSE echo from a voluntary client CLOSE that crossed it, and the server takes identical action on both branches — the code selected a log line. Meanwhile every client released before that change treats any code outside `NORMAL_CLOSURE`/`GOING_AWAY` as a head-of-line poison strike, escalating a routine demote to a `PROTOCOL_VIOLATION` terminal that quarantines the store-and-forward slot holding the un-acked rows. That trades data on mixed-version fleets for a diagnostic. A distinguishable role-change code needs a negotiated capability with `NORMAL_CLOSURE` as the fallback; the design doc now says so explicitly rather than leaving the absence unexplained.

Net effect on this PR's diff: zero. The commits remain in history because the branch is squash-merged.

## Test plan

- `CursorWebSocketSendLoopPoisonFrameTest`, `CursorWebSocketSendLoopErrorClassificationTest` — 25 tests, green.
- `CursorSendEngineSlotReacquisitionTest`, `CursorDeferredCloseCallerTest`.
- `MmapSegmentRecoveryFaultTest`, `SegmentRingEdgeRecoveryTest`, `SegmentRingRecoveryGenerationTest`, `SegmentRingTest`.
- Full core suite including `QuestDBImplCloseLifecycleTest`; repeat both acknowledged-interrupt cases 15× pinned to one CPU; verify a deliberately restarted deadline fails the regression assertion.
- End-to-end: `SqlFailoverQwpDeferredCloseExactlyOnceTest` from questdb/questdb-enterprise#1113 with questdb/questdb#7366 pinned to this branch.


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 16 / 16 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/impl/QueryClientPool.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/client/impl/SenderPool.java|8|8|100.00%|

bluestreak01:
## Review — tandem level 3 (client half)

Reviewed together with questdb/questdb#7366 and questdb/questdb-enterprise#1113. All three trees at PR heads; merge base `9ccfadbe`.

**Verdict: approve with comments** — no blocking item. Gated only by coupling: questdb/questdb#7366 has one open Critical and pins this branch.

---

### What this PR actually changes

Worth stating plainly, because the body does not match the diff. Against the merge base this is **four files**:

- `QueryClientPool.java` +18 and `SenderPool.java` +18 — the same change in both: a `private volatile Runnable creationWaitRetryHook` test seam, invoked in `close()`'s creation-wait loop after an `InterruptedException` when the loop will retry, plus a `@TestOnly` setter.
- `QuestDBImplCloseLifecycleTest.java` (+70/−80) — reworks two interrupt tests onto that seam.
- `design/qwp-nack-policy-v2.md` (+7/−2).

The "Store-and-forward recovery correctness" and "Shutdown and lifecycle" sections describe ~10 fixes — SIGBUS on recovered sparse segments, fail-closed on unreadable segments, the double-scan removal, O(S) `drainTrimmable`, the worker-quiescence barrier, the bounded deferred-close daemon — that are **in neither this diff nor the OSS diff**. They already landed on `main` via #64, #67 and #72 and are inherited through the merge base; `git diff origin/main HEAD` is identical to the PR diff, which is the proof. Located on `main`: `MmapSegmentRecoveryFaultTest` ← #64, `CursorSendEngineSlotReacquisitionTest` ← #67, O(S) `drainTrimmable` at `SegmentRing.java:1175`.

The "Close-code handling — added then reverted" section is the one part of the three PR bodies that is accurate about the wire contract. Both the enterprise and OSS bodies still describe ROLE_CHANGE (4001) as shipping; it is not. Please trim the inherited-work sections to what this branch actually carries — release notes drafted from this body would credit the wrong PR.

---

### Moderate

**M1 — The hook runs arbitrary caller code under the pool lock, unguarded** (`SenderPool.java:1504-1509`, `QueryClientPool.java:354-359`)

The creation-wait loop runs inside `lock.lock()` (`SenderPool.java:176`, released in the `finally` at `:1556`), and `creationFinished` is that lock's Condition — so `hook.run()` executes in the critical section. Two hazards, neither reachable in production (the field is null unless the `@TestOnly` setter is called, and the shipped hooks are non-blocking `Semaphore.release`):

- A hook that **blocks** on a thread that needs the pool lock deadlocks `close()` with no timeout — `ReentrantLock.lock()` is uninterruptible, so even the interrupt storm the tests drive cannot break it.
- A hook that **throws** propagates out of `close()` *after* `closeStarted = true` (`:1476`). A retried `close()` then returns immediately at the `if (closeStarted) return;` guard (`:1473-1475`), so the idle-delegate teardown loop after `:1558` never runs — permanently leaking every idle delegate (flock + mmap'd ring + IO thread).

This faithfully mirrors the pre-existing `borrowWaitExpiredHook` seam (`:222-225`, `:1410-1416`), and `SenderPool` already carries 24 `@TestOnly` members — so the seam itself is an established pattern, not new weight. The gap is that neither seam documents its contract. Worth: a `try/finally` (or `try/catch`) around `hook.run()`, and one line on the `@TestOnly` setter stating "must not block, must not throw, must not re-enter the pool".

**M2 — The reworked interrupt test retains a narrow window where its key assertion can pass vacuously** (`QuestDBImplCloseLifecycleTest.java:212-217`, helper at `:583-600`)

The closer reads its own interrupt status **inside** `try`, but `isCloseComplete` publishes in `finally`:

```java
try {
    db.close();
    isCloseReturnedInterrupted.set(Thread.currentThread().isInterrupted());   // read here
} finally {
    isCloseComplete.set(true);                                                // published here
    closeProgress.release();
}
```

`interruptUntilClose` interrupts whenever it takes a permit and `isCloseComplete` is still false. The ordering *retry-hook releases permit → budget expires → `db.close()` returns → test acquires permit, sees `isCloseComplete == false`, interrupts → status read* is not excluded. In that interleaving the assertion at `:238-239` observes the **test's own** interrupt, not the pool's `Thread.currentThread().interrupt()` restore — so it would still pass with the production restore deleted.

Strictly better than base, where a free-running `while (keepInterrupting.get()) { closer.interrupt(); Thread.yield(); }` thread made the same assertion vacuous over a far wider window. This PR narrowed it; it did not close it. Fix: gate the interrupt on the pool still being inside its wait, using the accessor this file already calls in `awaitCreationWaiter` (`:511`, `:522`):

```java
if (!isCloseComplete.get() && pool.hasCreationWaiterForTesting()) {
    closer.interrupt();
}
```

---

### Verified correct

- **The hook is tested in both pools** — installed at test lines 200 and 286. A one-sided-coverage suspicion was checked and did not hold.
- **The interrupt handling is byte-identical to base** (`git show 9ccfadbee2ac:…SenderPool.java:1486-1493`). A hypothesis that the un-cleared interrupt flag would busy-spin `awaitNanos` until the deadline was checked and rejected: `AQS.ConditionObject.checkInterruptWhileWaiting` uses `Thread.interrupted()`, which clears the flag before throwing.
- `isRetryingAfterInterrupt` is correctly scoped per iteration; `creationWaitInterrupted` still latches for the final `Thread.currentThread().interrupt()`; the hook fires only when the loop will genuinely retry (`inFlightCreations > 0 && creationRemainingNanos > 0`).
- **The rework is a net simplification** — two threads and ~28 lines of scheduler-dependent scaffolding replaced by an acknowledged handshake; the shared `interruptUntilClose` helper is used by both tests and alphabetically placed. The renames (`closeReturnedInterrupted` → `isCloseReturnedInterrupted`, `interruptCount` → `interruptRetryCount`, `keepInterrupting` dropped) are the right ones.
- **`WebSocketClient` does echo CLOSE** — `:1215-1231` calls `sendCloseFrameEcho(closeCode)` on `WebSocketOpcode.CLOSE`, before `upgraded = false` and before `handler.onClose(...)`. This is the premise the entire server-side close-echo handshake rests on, so it was verified directly rather than assumed. It holds. **But** it holds by code inspection, not by any test — see C1 on questdb/questdb#7366, which is the one blocking item across the three PRs.
- No new reflection in the diff; the `java.lang.reflect.Proxy` import in the lifecycle test pre-dates this PR.

---

### Minor

`WebSocketClient.java:1216` defaults `int closeCode = 0` when `payloadLen < 2`, so a server CLOSE with no payload makes the client echo close code `0`, which is not a legal wire code. Irrelevant during the server's echo wait (the code is discarded there), but the non-echo paths do parse and normalise an inbound close code. Not investigated further — noting it as a thread worth pulling separately.

---

### Summary

- **Verdict: approve with comments.** No Critical here.
- **Test gate: passes.** The seam is covered in both pools; M2 is an assertion-strength note, not a coverage gap.
- Before merge: trim the body to the work this branch carries (the SF-recovery sections belong to #64/#67/#72), and consider the two-line hardening on the hook contract (M1).

<sub>Review performed at level 3 with a 13-agent fan-out across all three repos; findings verified against source, with the regression set and the two candidate blockers verified empirically.</sub>

PR #76, opened 2026-07-28
feat(qwp): add table options API to name the designated timestamp column

jerrinot:
Sender users could not control the designated timestamp column name of
auto-created tables, QWP servers always created `timestamp`. This PR adds a
table-options API to set it:

```java
sender.table("trades").tableOptions().designatedTimestamp("ts");
```

## Changes

- New `Sender.TableOptions` interface, obtained via `tableOptions()` after
  `table()`. Options are namespaced so future create-time hints
  (partitioning, storage policies) land beside this one instead of on
  `Sender` directly. ILP senders reject the call — ILP cannot express it.
- Both QWP transports (WebSocket, UDP) encode the hint in the new per-table
  options trailer. When no hint is set, the emitted bytes are identical to
  the previous encoding, pinned by a golden-bytes test.
- The WebSocket client reads the server's `X-QWP-Table-Options` capability
  header and logs one warning when the hint will not be honored.

## Limitations

- Create-only hint: existing tables and pre-support servers ignore it.
- The returned options object is a reused per-sender instance; retaining it
  across table selections is documented as unsupported.
- The hint can be changed only while a table has no buffered rows.

Tandem PR: https://github.com/questdb/questdb/pull/7435

### DIFF
--- a/core/src/main/java/io/questdb/client/Sender.java
+++ b/core/src/main/java/io/questdb/client/Sender.java
@@ -743,6 +743,25 @@ public interface Sender extends Closeable, ArraySender<Sender> {
     Sender table(CharSequence table);
 
     /**
+     * Returns create-time options for the currently selected table.
+     * <p>
+     * Call this after {@link #table(CharSequence)}. QWP senders retain options
+     * per table for the sender's lifetime. ILP transports cannot express table
+     * options.
+     * <p>
+     * The returned instance is a single per-sender object that this method and
+     * {@link #table(CharSequence)} re-target at the currently selected table.
+     * Do not retain the reference across table selections; use it immediately
+     * and obtain a fresh one after selecting a table.
+     *
+     * @return options for the currently selected table
+     * @throws LineSenderException if the transport is ILP or no table is selected
+     */
+    default TableOptions tableOptions() {
+        throw new LineSenderException("table options are not supported by ILP");
+    }
+
+    /**
      * Add a column with a non-designated timestamp value.
      *
      * @param name  name of the column
@@ -835,6 +854,32 @@ public interface Sender extends Closeable, ArraySender<Sender> {
         FLUSH,
         APPEND,
         PERIODIC
+    }
+
+    /**
+     * Create-time options for the table selected by {@link Sender#table(CharSequence)}.
+     */
+    interface TableOptions {
+
+        /**
+         * Sets the designated timestamp column name to use if the selected
+         * table is auto-created by QWP ingestion.
+         * <p>
+         * Supporting servers use this hint only while creating a missing table
+         * and silently ignore it for existing tables. Older QWP servers also
+         * silently ignore it and create the conventional {@code timestamp}
+         * column instead.
+         * <p>
+         * The hint is sticky per table. It may be re-declared with a different
+         * name while the table has no buffered rows; changing it while rows
+         * are buffered throws.
+         *
+         * @param columnName non-empty column name of at most 127 UTF-8 bytes
+         * @return this instance for method chaining
+         * @throws LineSenderException if the name is invalid or these options
+         *                             no longer refer to the selected table
+         */
+        TableOptions designatedTimestamp(CharSequence columnName);
     }
 
     /**
--- a/core/src/main/java/io/questdb/client/Sender.java
+++ b/core/src/main/java/io/questdb/client/Sender.java
@@ -870,9 +870,9 @@ public interface Sender extends Closeable, ArraySender<Sender> {
          * silently ignore it and create the conventional {@code timestamp}
          * column instead.
          * <p>
-         * The hint is sticky per table. It may be re-declared with a different
-         * name while the table has no buffered rows; changing it while rows
-         * are buffered throws.
+         * The hint is sticky per table. It may be set or re-declared only
+         * while the table has no buffered rows; setting it while rows are
+         * buffered throws.
          *
          * @param columnName non-empty column name of at most 127 UTF-8 bytes
          * @return this instance for method chaining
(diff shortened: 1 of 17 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 135 / 138 (97.83%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|21|23|91.30%|
|:large_blue_circle:|io/questdb/client/cutlass/http/client/WebSocketClient.java|23|24|95.83%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|8|8|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketEncoder.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpUdpSender.java|30|30|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpColumnWriter.java|18|18|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/protocol/QwpTableBuffer.java|16|16|100.00%|
|:large_blue_circle:|io/questdb/client/Sender.java|1|1|100.00%|

jerrinot:
M2/M3 confirmed against the diff source (`setDesignatedTimestampName`: the `else if (!Chars.equals(...))` path skips the `rowCount != 0` throw, so a same-name re-set while rows are buffered is a silent no-op — contradicting the javadoc's "any set while rows are buffered throws"). Here's the review.

---

# Code Review: PR #76 — `feat(qwp): add table options API to name the designated timestamp column`

**Reviewed at level 3** (full mission-critical pass): change-surface map, self-verification of all load-bearing wire-format/estimate logic, plus three parallel bug-hunters (cross-context callers, concurrency/resource/pool, fresh-context adversarial — the last cross-checked the client bytes against the parent-repo server decoder `QwpMessageCursor.parseTableOptions`).

**Committed-binary gate: clean** — no binary files in the diff.

This is a careful, well-tested PR. **No critical or blocking defect.** Every high-risk vector — trailer byte arithmetic, the UDP datagram estimate, store-and-forward replay, flag-bit allocation, header parsing, concurrency — holds up under byte-level verification against the golden-byte tests.

## Moderate

**M1 — The client emits `FLAG_TABLE_OPTIONS` + trailer unconditionally, even to servers that never advertised support. Confirm the server contract before merging ahead of the tandem PR.** *(in-diff, by-design, external dependency)*
`QwpUdpSender.sendEncodedPayload`/`appendTableOptionsForUdp` and `QwpWebSocketEncoder.finishMessage` set the flag and append the trailer whenever a name is declared, without gating on the negotiated capability — `warnIfTableOptionsUnsupported` only logs. This is deliberate and *correct for the client*: store-and-forward frames must replay to arbitrary servers, so gating on a per-connection handshake is wrong. `QwpConstants.java:54-64` documents the reliance on the server forward-compat guarantee: pre-support servers ignore unknown flag bit `0x20` and never read past the last table body (the trailer rides inside the existing `payloadLength` envelope). The fresh-context agent verified the *current* server decoder agrees byte-for-byte; the unverifiable part is a genuinely *old* server. Nothing to change here — just don't merge the client ahead of a verified server contract, and consider a mixed-version e2e test. If the guarantee holds, an old server silently creates `timestamp` and the log warning is the only signal; if it doesn't, frames get rejected.

## Minor

**M2 — Re-declaring the *same* name while rows are buffered is a silent no-op, contradicting the javadoc.** *(in-diff)*
`QwpTableBuffer.setDesignatedTimestampName`: the `else if (!Chars.equals(designatedTimestampName, columnName))` guard means an identical-name re-set skips both the `rowCount != 0` check and the throw. Both the method javadoc and `Sender.TableOptions.designatedTimestamp` state "setting it while rows are buffered throws." Behavior is benign (idempotent; trailer size unchanged; the UDP base-estimate cache is only read at `rowCount == 0`), but the doc/behavior mismatch is real and untested. Fix: move the `rowCount != 0` guard ahead of the equals-check, or soften the javadoc to "*changing* it while rows are buffered throws." One-line fix.

**M3 — Stale explanatory comment in `QwpUdpSender.TableOptionsImpl.designatedTimestamp`.** *(in-diff)*
The comment claims "Any accepted set requires rowCount == 0" — false for the M2 same-name path (accepted with `rowCount > 0`). The conclusion (no `committedDatagramEstimate` adjustment needed) still holds, but because a same-name set can't change the trailer size, not for the stated reason. Fix alongside M2.

## Downgraded (verified false positives)

- **UTF-8 length mismatch (validation vs encoding)** — `Utf8s.utf8Bytes` (validation) and `NativeBufferWriter.utf8Length` are the same function, also what `putString` writes. Byte-exact.
- **`reset()` wiping the sticky name** — the null-assignment is in `clear()`, not `reset()`; sticky-across-flush contract holds (verified directly).
- **`FLAG_TABLE_OPTIONS = 0x20` collision** — free bit (existing `0x01`, `0x04`, `0x08`, `0x10`).
- **Header parser matching `"10"`/`"21"` as tag `1`** — exact `"1".equals(token.trim())`, not a substring test; no infinite loop / OOB; reset to `false` before each upgrade.
- **`payloadStart` promoted to a field** — encoder is single-threaded, each entry point re-stamps it; pinned by the byte-identical golden test.
- **UDP estimate under-counting the trailer** — `getSingleTableOptionsTrailerSize` is byte-exact vs `writeTableOptions`+`putInt`, folded into the base estimate and invalidated on every rename; `estimate >= actual` holds.
- **Concurrency / pool staleness / leak** — capability volatiles gate only logging, never wire bytes; `PooledTableOptions` guards generation on every call; flyweights hold no native memory.
- **Missing `tableOptions()` override** — all three QWP senders override; ILP (`AbstractLineHttpSender`) inherits the throwing default; `PooledSender` delegates with a generation guard.

## Test coverage

Strong. Golden-byte assertions on both transports, sticky/reset semantics, stale/retained-reference retargeting, estimate invalidation on rename, ILP rejection, header parser (present/absent/other-tag), mixed-presence and consecutive-split-message trailers, pool generation guard. The reflection in `WebSocketClientTest`/`SenderLeaseGenerationTest` reuses the file's existing pattern for package-private internals — acceptable. Only gap: M2's no-op path is untested (which is why the doc drift slipped through).

## Summary

- **Verdict: approve.** No blocking issues. M2/M3 are one-line doc/comment fixes worth folding into this branch; M1 is a cross-repo release-ordering check, not a code change.
- **3 findings verified, 8 dropped as false positives** after source verification.
- **Split: 3 in-diff, 0 out-of-diff.** Zero out-of-diff is a genuine result, not an underrun — the feature is additive and the trailer travels inside the existing payload-length envelope; every encoder/UDP-send caller (`flushPendingRows`, `flushPendingRowsSplit`, `sendCommitMessage`, all UDP paths) was traced and is safe.

PR #79, opened 2026-07-31
test(qwp): deflake close-lifecycle interrupt tests

mtopolnik:
## What

Fixes flakiness in the two `QuestDBImplCloseLifecycleTest` tests that check `close()` stays bounded under repeated interrupts (the query and the sender variant). On a busy CI machine they could fail with:

```
close left its creation wait before the interrupt storm landed twice; interrupts landed: 1
```

even though the product code honored its deadline exactly as intended.

## Why it flaked

Each test used a background thread to interrupt the closing thread, then waited to observe at least two interrupts land while `close()` was still waiting. That wait read `hasCreationWaiterForTesting()`, which briefly reports `false` every time an interrupt wakes the waiting thread before it goes back to sleep. On a busy machine the background thread was slow to run, and the wait caught that brief gap after a single interrupt, so it failed while the product was correct.

A previous change raised the close budget from 100ms to 1s, which made this rarer but did not remove the race.

## Fix

- The test thread now delivers one interrupt itself while `close()` is definitely still waiting. This guarantees an interrupt lands during the wait, so the test's existing "close restored the interrupt flag" assertion is reliable (the product restores that flag only when it caught an interrupt during the wait).
- A background thread keeps interrupting through the join. That is the part that would actually catch a real bug: if an interrupt reset the deadline, `close()` would never return and the join would time out.
- Removed the fragile "wait for two interrupts to land" helper and a redundant counter assertion.

Behavior of the product code is unchanged; this is a test-only change.

## Test plan

- `mvn -pl core test -Dtest=QuestDBImplCloseLifecycleTest` passes (6/6).
- Ran the two interrupt tests 5 extra times back to back, green each time, ~2.3s per run.

### DIFF
--- a/core/src/test/java/io/questdb/client/test/impl/QuestDBImplCloseLifecycleTest.java
+++ b/core/src/test/java/io/questdb/client/test/impl/QuestDBImplCloseLifecycleTest.java
@@ -43,7 +43,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.function.IntFunction;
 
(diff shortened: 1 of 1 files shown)


jovfer (approved):

mtopolnik:
PR no longer relevant

PR #91, opened 2026-08-19
feat(qwp): recycle the symbol dictionary at a threshold to lift the cardinality cap

### LINKED ISSUE #80: 1 million of district symbols limit
### To reproduce

Review point M6 in https://github.com/questdb/java-questdb-client/pull/66#issuecomment-5183155169 is very valid

M6 — At the 1M cap, the error message's remedy is not performable through SenderPool (out-of-diff). getOrAddSymbol says "close this sender and build a new one", but PooledSender.close() (:153) gives the delegate back rather than closing it, and reset() does not clear globalSymbolDictionary. A delegate that hit the cap is handed to the next borrower still full, throwing on every new symbol value. Only incidental reapIdle recovers it.

### QuestDB version:

latest

any

any

### Full Name:

Alex Pelagenko

### Affiliation:

QuestDB

- [x] Yes, I have

### Additional context

_No response_

jovfer:
Fixes #80

The WebSocket sender no longer has a hard symbol-cardinality stop. When its global symbol dictionary reaches a configurable threshold (`symbol_dict_reset_threshold`, default 100 000), the sender recycles its send stack at the next provably-safe instant — all published frames acknowledged, no row in progress, no open transaction: it closes the send loop, retires the store-and-forward slot with the existing fully-drained close (emptying it), starts a fresh dictionary epoch on the same slot, rebuilds the engine through the same code path `build()` uses, and reconnects. The server needs no change: supported servers (10.x+) already clear per-connection dictionary state on disconnect, and this branch pins that behavior as a wire contract. Zero wire change, zero on-disk format change; all production code is client-side.

## User-visible surface

Three connect-string keys / builder methods — `symbol_dict_reset` (default on), `symbol_dict_reset_threshold`, `symbol_dict_reset_max_wait_millis` (bounded wait for a starved reset; default `2000`, at most one wait per armed window; `0` makes the recycle opportunistic-only); an advisory `Sender.resetSymbolDictionary()` (no-op on non-QWP transports; producer-thread-only, and `PooledSender` forwards it to the live delegate); observability getters on `QwpWebSocketSender` (`getSymbolDictEpoch`, `getSymbolDictResetStarvationTimeouts`, `isResetArmed`, `isSymbolDictResetEnabled`, `getSymbolDictResetThreshold`, `getSymbolDictResetMaxWaitMillis`). Externally visible frame sequence numbers stay strictly monotone across recycles via an epoch-base offset applied on every public surface (flush sequence, ack watermark, drain, progress stream, error spans). The client-side dictionary cap rises 1M → 2M (now equal to the server cap, with an OSS ordering test), and the cap error now points at the reset valve. A recycle also naturally heals a mid-life full-dict degrade (the fresh engine re-derives delta capability). Senders built via the public `connect(...)` overloads (no rebuild factory) and senders that don't own their engine never recycle and never arm, so `isResetArmed()` cannot read true against a permanently-zero resets counter.

Release-note items:

- The `CursorWebSocketSendLoop` public constructors gained a parameter in place. `QwpWebSocketSender.connect(...)` is additive: a new overload carries the recycle settings and every pre-existing overload is unchanged.
- The per-connection counters (`getTotalFramesSent`, `getTotalFramesReplayed`, `getTotalReconnectAttempts`, `getTotalReconnectsSucceeded`, `getTotalServerErrors`, `getTotalAcks`, `getTotalBackpressureStalls`) are sender-lifetime: the sender owns one counter holder that every rebuilt I/O loop and every attached engine adopts before use, so the values carry across recycles and are retained after `close()`. A recycle's own reconnect counts as one attempt and one success.
- Servers released before QuestDB 10.0.0 cap the symbol dictionary at 1M, and QWP has no wire-level negotiation of the limit. Only 10.0.0+ servers are supported; note that the re-arm floor lets a default sender's dictionary grow to 1M entries per epoch, so this is not an edge case of exotic configuration.
- `getAckedFsn()`'s contract is restated on the `Sender` interface: on a live sender it never collapses back to −1 and keeps reporting the last durable watermark across a recycle.
- `markEverConnected()` is a new public method on `CursorWebSocketSendLoop` (loop-side seeding of the sender-lifetime ever-connected flag; `QwpWebSocketSender` calls it when it rebuilds the loop).
- `symbol_dict_reset_threshold` accepts 1..1,000,000 (half the protocol cap, the re-arm floor's own ceiling); `symbol_dict_reset_max_wait_millis` rejects values whose nanosecond conversion overflows.
- `getSymbolDictResetsPerformed()` is gone — it could never differ from `getSymbolDictEpoch()`.

## Recycle robustness

The through-line: the recycle prefers abandoning and resuming over latching or re-arming.

- **Reconnect is deferred to the I/O loop.** The post-swap reconnect never re-runs the initial-connect policy on the producer thread: `ensureConnected()` latches `hasInitialConnectRun` on its first completion and routes every later entry through the ASYNC branch — the loop is built and started synchronously, the socket connect happens on the I/O thread with indefinite retry, and the producer keeps buffering into the fresh epoch's slot. A default (OFF) or SYNC store-and-forward sender therefore never sees a transport error or a reconnect budget past its initial connect, matching the steady-state store-and-forward contract. The barrier swaps only while the I/O loop holds a live connection (`CursorWebSocketSendLoop.isLinkUp()`): step 2's loop close cancels a live socket in milliseconds, but a reconnect blocked in a hostname resolve or a credential pull ignores the cancel and would hold `table()` for the join budget, so while the loop is between connections the recycle stays armed and runs at the first drained barrier after the reconnect; `SymbolDictRecycleOutageTest` pins both halves. Deferring has a bound: the dictionary keeps growing until the reconnect, and the protocol caps a sender at 2,000,000 distinct symbol values (`MAX_SYMBOL_DICTIONARY_SIZE`), so a producer that registers that many new values above the threshold within one outage fails `symbol()` with the dictionary-full error instead of swapping mid-outage. The rebuilt loop is seeded with the sender-lifetime ever-connected flag (`markEverConnected`, read after the outgoing loop's `close()` has joined the I/O thread), so `wasEverConnected()` stays sticky and endpoint-policy failures keep their post-init retry classification across the swap.
- **Anti-thrash re-arm floor.** A live symbol set larger than the threshold would otherwise re-arm on every refill, recycling (teardown + reconnect + dictionary re-ship) every ~`L·ln(L/(L−T))` rows forever. Each swap raises the re-arm bar to `max(threshold, 2 × size-at-swap)` and never lowers it -- a manual `resetSymbolDictionary()` swap bypasses the floor for that one swap without resetting it --, capped at half the protocol cap so genuinely unbounded-cardinality producers keep recycling. No new config key; the default threshold is unchanged. Red-proofed: without the floor a bounded 6-symbol set recycles 4×, with it once. The floor arithmetic itself is pinned (doubling per swap, organic re-arm refused below it, advisory request bypasses it) — the integration suites' bounded symbol sets rely on that bypass to keep recycling.
- **A deferred engine close is awaited.** The fully-drained close can return with the slot flock retained when the SF worker is wedged in a syscall past `SegmentManager`'s bounded join (a stalled disk/NFS — exactly what the deferred-close machinery exists to survive). `awaitDeferredEngineClose` parks until the worker's exit path confirms the release (30 s budget) instead of letting the rebuild throw `SlotLockContentionException`. `SymbolDictRecycleDeferredCloseTest` pins both branches with a wedged-worker harness and a positive parked witness; red-proofed against the pre-fix code (dies with the exact `SlotLockContentionException` trace). The same protocol closes an engine the rebuild cannot keep (recovered fully-acked leftovers): a wedged worker there retains the engine and resumes on the next send instead of colliding with the next rebuild, and `close()` keeps reporting the flock held until the worker exits; pinned with a wedged recovered engine both riding the stall out and exhausting the await. Step 3 detaches the outgoing engine's slot-lock listener before closing it, so a release completing after the swap cannot mark the rebuilt engine's flock released; a recycle-retained engine reaches pools through the `isSlotLockReleased()` re-probe. The await budget is spent once per pending close: a resume while the worker is still wedged probes and rethrows without parking (pinned: the retried `table()` returns in well under the budget).
- **`build()` never throws for an error the I/O thread latched during construction.** The rebuild-factory setter refuses only a closed sender (a rejection latched while `build()` runs surfaces on the first send, as before, and `close()` cleans up), and both post-connect setters sit inside `build()`'s close-on-failure guard, so a throw there can no longer leak the connected sender, its threads and its slot lock. This was the Windows CI failure of `SenderPoolDataLossNotificationTest` on the pool's startup recovery.
- **The recycle is resumable instead of terminal.** The swap commits only after a fresh engine stands on the emptied slot (dictionary, epoch counters, and wiring move after the rebuild), and every transient mid-recycle failure — an interrupt arriving while the loop close joins the I/O thread (a flag the producer carries in is cleared for the join and handed back, so it never abandons), a wedged SF worker exhausting the deferred-close budget, a momentary EMFILE at rebuild, a post-cleanup fsync warning — records a two-state resume point (`CLOSE_LOOP`/`REBUILD`) that the next send completes. The terminal latch survives for exactly one case: a rebuilt engine that recovered *unacknowledged* frames, which proves the outgoing close's fully-drained contract was breached. Benign recovered fully-acked leftovers (a transiently failed segment unlink) heal by closing the recovered engine — which retries the unlink by design — and rebuilding. `Error` passes through the recycle machinery unwrapped and unlatched. Suites pin the interrupt, rebuild-fault, await-exhaustion, heal, and breach paths, each red-proofed.
- **A rebuild-time quarantine reaches the handler installed on the sender.** The rebuild factory receives the sender's current user handler (`rebuild(SenderErrorHandler)`), not the builder's field captured at build time, so `setErrorHandler()` after `build()` is honoured when a recycle rebuild sets a torn slot aside; the builder's factory also snapshots `sf_dir`/sender id at build time.
- **A failed post-swap reconnect never latches.** A `@TestOnly` fault hook drives a real failed reconnect: one test proves ids registered mid-window survive into the flushed delta (fails with the `resetSymbolDictStateForNewConnection` guard reverted to the unconditional clear), the other proves the failure leaves the sender usable (fails with `recycleFailure` latched in the reconnect catch). Both counterfactuals re-run red against the final tree.
- **`close()`'s drain guard is split by true dependency** (flush/commit/seal need only the engine; `checkUnsurfacedError`/`drainOnClose` need the loop). Staged rows cannot coexist with a null loop — `sendRow()` calls `ensureConnected()` before a row commits and every failure path rolls the in-progress row back — so this is hardening that keeps the guard honest for any future state, not a fix for a reachable row drop.
- **Accessors are truthful across the recycle window.** `getAckedFsn()`/`awaitAckedFsn()` snapshot `cursorEngine`/`cursorSendLoop` into locals and, while the engine is null (mid-swap), report the durable watermark the recycle barrier proved instead of collapsing to −1; the epoch base is read before the engine so a torn pair can never fabricate an FSN above anything published; `cursorEngine`, `cursorSendLoop`, `fsnEpochBase`, `lastRecycleDurableFsn` and the ever-connected flag are volatile for monitoring readers, and the seven `getTotal*` counters read one sender-lifetime holder; the `getAckedFsn` javadocs state the real post-recycle sentinel behavior, including the HTTP/TCP/UDP always-−1 caveat. The torn (old base, fresh engine) clamp is pinned by a test that parks a monitor thread between the accessor's two reads across a real recycle; the null-engine window is asserted from inside the deferred-close park witness — both go red with the clamp or the durable write removed.
- **Delta-dict healing is epoch-scoped by design.** The rebuilt engine re-derives delta capability rather than inheriting a one-way degrade latch: `SymbolDictRecycleHealingTest` pins that a degraded sender heals back into delta mode once the fault clears, and that a persistent fault degrades the fresh engine again as a normal, catchable error.
- A rebuild that recovers *unacknowledged* frames latches the sender terminal (empties-the-slot contract breach); fully-acked leftovers heal. The stale slot-lock latch a pool-style re-probe can leave behind is cleared as soon as the outgoing close is confirmed, before the rebuild takes the flock (pinned through a rebuild-factory probe), and the fresh dictionary is allocated before the rebuild so the commit block genuinely cannot throw.
- The fresh dictionary is allocated at the outgoing dictionary's size (the recovery path already did this), so an epoch's refill no longer re-climbs the rehash ladder.
- **Diagnostics use the external FSN scale.** `drainOnClose`'s two WARNs (the abandon-uncommitted-frames one and the timeout one) and its thrown message, plus the I/O loop's two replay log lines, report epoch-rebased FSNs — the same scale `flushAndGetSequence()` and the `SenderError` spans use (pinned after a real recycle).
- **Contracts the recycle touches are stated where readers look.** A recycle's reconnect fires `RECONNECTED`/`FAILED_OVER` with no `DISCONNECTED` (now on `SenderConnectionEvent.Kind` and `SenderConnectionListener`, and pinned); a rebuild-time quarantine is delivered synchronously on the producer thread like the build-time one (`SenderErrorHandler`); `setEngineRebuildFactory` is documented and refuses a closed sender. `setEngineRebuildFactory(null)` stops future arming but never cancels an armed or pending recycle, and its javadoc now says so.

This update sets the recycle's default starvation wait to 2 s and closes two resume gaps found under interrupt and fault injection. `symbol_dict_reset_max_wait_millis` defaults to `2000`: a recycle armed that long without a `table()` call finding the backlog drained pauses the next `table()` call for at most 2 s so the backlog can drain, then swaps — at most once per armed window; `0` opts out to opportunistic-only. A continuous producer keeps frames in flight at every row start and never exposes a drained instant on its own, so the wait is what makes the recycle reachable for the population that crosses the threshold; on a healthy link the pause is about one ack round trip, because the paused producer stops refilling the ring. The value is sized against the sender pool's acquire timeout (see Design decisions) and pinned by `SymbolDictResetDefaultsTest`. The two park loops a recycle can enter (`maybeBlockForStarvedReset()`'s starvation wait, `awaitDeferredEngineClose()`'s deferred-close wait) and the loop-close join share one interrupt policy — clear the flag for the park or join and hand it back on every exit, so the recycle never consumes the caller's interrupt — after a carried interrupt flag was found manufacturing a false `CLOSE_LOOP` abandon out of an otherwise successful wait. A `CLOSE_LOOP` resume now re-registers the delta baseline it used to drop, republishing it as deferred dictionary chunks the fresh loop replays ahead of any data frame, and closes the outstanding chunk-commit debt on both the resume's `Throwable` and `Error` paths, so an interrupted or faulted recycle converges the same way a clean one does. `resetSymbolDictionary()`, the dictionary-cap error, and `armIfEligible()`'s javadoc now state the trigger point in one sentence each: `table(...)` is the only place a recycle ever starts; a recycle that failed transiently mid-swap is resumed and completed by the next `table()`/`flush()`/`drain()` call (an `at()`/`atNow()` that finds the swap still pending fails and rolls back its row, and the next `table()` completes it), and `SenderErrorHandler`'s threading contract says so too, since a rebuild-time quarantine can therefore arrive synchronously on the producer thread from any of those calls. Tests pin these: the default wait's sizing (positive, and at most half the pool's acquire timeout), the interrupt policy under both park loops, the CLOSE_LOOP re-registration and its fallback/debt-close paths (including a crash-replay run proving the abandoned resume group survives a restart), a post-recycle endpoint-policy-rejection retry, and FSN epoch-base accumulation across two real recycles.

## Design decisions

- **A 2 s bounded wait by default** (`symbol_dict_reset_max_wait_millis=2000`): the recycle runs when a
  `table()` call finds the backlog drained; a recycle armed for 2 s without one pauses the next `table()`
  call for at most 2 s so the backlog can drain, at most once per armed window. A continuous producer
  keeps frames in flight at every row start and never exposes a drained instant on its own, so an
  opportunistic-only default would leave the recycle unreachable for exactly the population that crosses
  the threshold; with the producer paused, a healthy link drains in about one ack round trip. The wait
  surfaces no transport error, carries no reconnect budget and never latches: on timeout the call
  returns, the sender stays armed, and no second wait runs until a swap commits, so a drop that lands mid-wait
  costs at most the rest of that one pause, and a loop already between connections when the barrier
  runs is skipped outright (the live-connection gate above). The earlier 30 s default failed on pool arithmetic: a give-back flush arms the sender,
  the slot idles past the window (`idle_timeout_ms` 60 s), and the next borrower paid the wait while
  holding its lease, longer than `acquire_timeout_ms` (5 s), so every other acquirer timed out. 2 s
  clears that with margin, covers one round trip plus a modest backlog, and survives the first four
  attempts of the reconnect backoff ladder. A producer that outruns the wire still times out and stays
  armed until a natural gap; the signal is `getSymbolDictResetStarvationTimeouts()` at 1 with
  `isResetArmed()` true while `getSymbolDictEpoch()` does not advance, and the planned escalation for
  that population is an asynchronous handover of the dictionary swap to the I/O loop, not a longer
  wait.
- **Recycle bounds mirror `close()`'s stall class, not always its blocking behavior.** The swap's
  loop-close wait is literally `close()`'s own budget (`CursorWebSocketSendLoop.close()`'s 30 s
  shutdown-await) — the same method a plain `close()` calls. The deferred-engine-close wait is a
  separate, dedicated budget (`RECYCLE_DEFERRED_CLOSE_MAX_WAIT_MILLIS`, also 30 s) sized to survive the
  same disk/NFS stall class `SegmentManager`'s bounded worker join is built to survive; unlike a plain
  `close()`, which spends only its own join budget and returns, leaving eventual release to a pool's
  background re-probe (`isSlotLockReleased()`), the recycle parks the producer on this wait so the
  rebuild never collides with a retained flock. Rows are refused only for that window, and only while
  the SF worker is genuinely wedged — a durability-failure class, the same kind of signal
  `SegmentRing.checkDurability()` already surfaces on the row-send path — never on a transport outage,
  and no reconnect budget applies to the producer. Both parks' interrupt policies are pinned: `SymbolDictRecycleStarvationTest` for the starvation wait, `SymbolDictRecycleDeferredCloseTest` for the deferred-close wait (an interrupt carried into either park is handed back on every exit; both tests fail against a mutated exit policy).
- **The CLOSE_LOOP resume's re-registration has its own producer-visible bound.** Republishing the
  dropped delta baseline runs through the same `sealAndSwapBuffer`/`appendBlocking` path as any other
  publish, on the producer thread, bounded by `sf_append_deadline_millis` (default 30 s) — stated as its
  own bound, like the deferred-engine-close wait. A failure mid-publish clamps the delta baseline to
  the chunks that reached the ring and closes their group best-effort; a failure after the publish
  keeps the baseline (coverage is full) and retries the closing commit (`commitOrphanedDictionaryChunks`,
  which never throws); if that best-effort close still leaves the group open, every subsequent `flush()`
  and `close()` re-attempts the closing commit whenever a group is open (`hasDeferredMessages`), so the
  residual is a retried, not permanent, state. The re-registration is one-shot, so a remainder wider than
  the server cap after a mid-publish failure stays wedged on the split pre-flight -- a narrower
  population than the former baseline drop, which re-shipped the whole range.
- **Server support floor is 10.x.** QWP was experimental through the 9.x line and never released with a
  9.x server, so no compatibility shim for 9.x server behaviour exists or is argued from.
- **`Sender` stays single-threaded.** A concurrent `close()` while a recycle waits is outside the
  documented contract (`Sender`'s own thread-safety javadoc).
- **The trigger point is `table()`, by contract.** `Sender.resetSymbolDictionary()`, the dictionary-cap
  error message, and `armIfEligible()`'s javadoc all state the trigger explicitly: the recycle runs at
  a `table()` call that finds the backlog acknowledged, and a caller that never starts another row never
  recycles. No per-row branch is added to the hot path.
- **Pooled senders inherit an armed recycle at give-back.** On a healthy link the ring drains during the
  idle window, so the next borrower's first `table()` swaps without waiting; only an unhealthy link makes
  it pay the bounded wait while holding its lease, which is why the default stays at half the pool's
  `acquire_timeout_ms`. An abandon during a give-back flush ejects the slot via `discardBroken`, which removes it from the pool's slot
  set so the pool can grow back into a fresh one on demand — safe, at a capacity cost until it does. The
  pooled-path recycle has no end-to-end coverage yet.

## Tests

New client suites cover arming, FSN continuity, the swap itself, memory mode, starvation, refusals, outage/orphans, catch-up skip, crash windows, partial re-registration, healing/metrics, deferred close, slot heal/breach, reconnect fault injection, and the reset defaults. Alongside them: the OSS disconnect-clear contract pin, real-client e2e in SF and memory modes, a seeded recycle×reconnect fuzz, and ENT failover losslessness with recycles in flight. Verified green from a clean reactor build; the reconnect-fault counterfactuals re-run red at the final tree.

## Companion PRs

- OSS (contract pin + e2e + fuzz): questdb/questdb#7525
- Enterprise (failover losslessness): questdb/questdb-enterprise#1171

https://claude.ai/code/session_01K2Hw6TfgX9sBB2L8oSdaWn
https://claude.ai/code/session_01NAsaMbjHq9NAENT1cJWbdg

https://claude.ai/code/session_01EnhV4ss4WoGSkemvdJdbuv


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 251 / 271 (92.62%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|70|85|82.35%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|162|167|97.01%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|14|14|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 429 / 493 (87.02%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|77|92|83.70%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|328|377|87.00%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|18|18|100.00%|

jovfer:
# Tandem review — level 3

**Scope:** enterprise questdb/questdb-enterprise#1171 + OSS questdb/questdb#7525 + client #91. All three local heads matched their pushed PR heads at review time (ENT `524efc380`, OSS `97f8fcb60`, client `dd3e90465`).

**Submodule provenance:**
- `questdb`: `fef82615` → `97f8fcb6` — **OFF-DEFAULT** (OSS feature branch) → in scope
- `java-questdb-client`: `0b9b5766` → `dd3e9046` — **OFF-DEFAULT** (client feature branch) → in scope

ENT #1171 and OSS #7525 are test-only. All production code is in this repo.

## Moderate

### 1. `cursorSendLoop` lacks the `volatile` its sibling was given for exactly this reason — `QwpWebSocketSender.java:297` (in-diff)

> **Problem:** Monitoring accessors read a non-volatile field the recycle now reassigns.
> **Net impact:** Ops dashboards can show a dead epoch's frozen counters indefinitely.
> **Evidence:** Static, at `dd3e9046`: declarations at :296/:297, write sites 3→5, 7 accessors.

Line 296 is `private volatile CursorSendEngine cursorEngine;`, carrying a comment this PR adds: *"Volatile since the recycle started reassigning it (non-null -> null -> non-null on the producer thread): the monitoring accessors ... read it from a monitor thread."* Line 297's `cursorSendLoop` gets no such treatment, despite the recycle reassigning it identically — write sites go from 3 at base (effectively write-once for a live sender) to 5 at head, including the two new recycle sites `:5734` and `:5808`.

The same PR adds this javadoc to seven accessors that read `cursorSendLoop`: *"Reads the live cursor I/O loop, which a symbol-dictionary recycle rebuilds, so the count restarts at 0 on every recycle: a monitor differencing it across one sees a negative delta."* So the change itself documents them as monitor-facing across a recycle. `getTotalBackpressureStalls` got that same javadoc but reads `cursorEngine`, which *was* made volatile — the accessor family was treated inconsistently.

Base: at `981bdb02a` the field was effectively write-once for a live sender, so the plain read was benign. No runtime artifact was produced, which is why this is Moderate and not Critical — the consequence is observability only.

**Fix:** `private volatile CursorWebSocketSendLoop cursorSendLoop;`

### 2. The anti-thrash floor can be lowered through the manual valve, contradicting its own comment — `QwpWebSocketSender.java:5429` (in-diff)

> **Problem:** Manual reset lowers the re-arm floor the code says is never lowered.
> **Net impact:** Each manual reset re-triggers ~log2(liveSet/threshold) extra reconnects.
> **Evidence:** Static, at `dd3e9046`: comment :488 vs code :5429; 5 total occurrences, no `Math.max`.

Line 488 (added by this PR): *"Never lowered -- a shrunken working set simply stops arming."* Line 5429 is a plain assignment:

```java
resetFloorSymbols = Math.min(dictSizeAtSwap * 2, QwpConstants.MAX_SYMBOL_DICTIONARY_SIZE / 2);
```

Automatic swaps genuinely cannot lower it — `armIfEligible:5262` requires `size >= max(threshold, floor)`, so `2*dictSizeAtSwap >= floor`. But `:5263`'s `|| manualResetRequested` bypasses the size gate entirely, so `resetSymbolDictionary()` can swap at `dictSizeAtSwap < floor/2` and lower the floor. A producer with live set L >> threshold T ratchets the floor to ~2L; a scheduled manual reset while the current epoch holds only S << L symbols sets it to 2S, and the sender re-walks the whole doubling ladder — one full teardown/rebuild/reconnect per rung. That is the thrash the floor exists to prevent, reintroduced through the public valve this PR adds.

Calling the valve at a quiet point is the pattern `symbolDictResetMaxWaitMillis`'s own javadoc recommends. Base: `N/A — new surface`.

`SymbolDictRecycleArmingTest.testReArmFloorDoublesPerSwapAndBlocksOrganicReArm` walks right up to this — asserts floor 8 at dict size 4, calls `resetSymbolDictionary()`, asserts *"the advisory request bypasses the floor"* — then ends without completing that swap. All three floor assertions in the suite follow automatic swaps.

**Fix:** `resetFloorSymbols = Math.max(resetFloorSymbols, Math.min(dictSizeAtSwap * 2, MAX / 2));` — or, if lowering after an explicit manual reset is intended, amend the `:488` comment. One of the two is wrong.

### 3. CLOSE_LOOP resume's partial-publish commit is anchored above its own chunk coverage — `QwpWebSocketSender.java:5831` (in-diff)

> **Problem:** Orphan commit's delta start can exceed the chunks that actually landed.
> **Net impact:** If reached, sender latches terminal with a false "host crash" message.
> **Evidence:** Static, at `dd3e9046`: :5831 / :6279 / :4446 / :5952 / :6528; no artifact for reachability.

The chunk range and the group-closing commit come from two different sources:

- `:5831` `publishDictionaryChunks(cap, 0, sentMaxSymbolId)` — chunks start at id **0**
- `:6279` `catch (Throwable t) { if (anyChunkPublished) commitOrphanedDictionaryChunks(t); throw t; }`
- `:4446` → `sendCommitMessage()` → `:5952` `commitBaseline = symbolDeltaBaseline()` → `:6528` `deltaDictEnabled ? sentMaxSymbolId : -1`

They agree only when the publish completes (chunks reach M, commit says M+1, `sentDictCount` reaches M+1 — passes). On a partial publish chunks reach k < M, so the commit's delta start M+1 exceeds `sentDictCount` k+1 and `CursorWebSocketSendLoop:3545` calls `recordFatal`. The arm's own `deltaDictEnabled && cap > 0 && sentMaxSymbolId >= 0` guard is what makes `symbolDeltaBaseline()` return M rather than −1, so this is the only call site where the two can diverge.

At base, `publishDictionaryChunks`' sole caller was the `!deltaDictEnabled`-gated full-dict fallback, so the orphan commit always carried delta start 0 and could never exceed coverage — not pre-existing.

**What is not proved:** reaching the branch needs a mid-chunk append failure while the ~16-byte commit still fits. That is a runtime-shape claim and no execution artifact was produced, so severity is deliberately not promoted. The CLOSE_LOOP state itself is known-reachable (production seam plus a shipped test); only the partial publish is unproven.

The two shipped resume tests inject via `resumeCommitFaultForTesting`, which fires only *after* `publishDictionaryChunks` returns with full coverage — so the partial shape has no coverage and no fault seam.

Related, and solid without execution: the memory-mode safety comments at `:5850-5853` and `:5871-5874` justify the baseline drop with *"any post-reset redefinition is FIFO-ordered after these rowless chunk frames"*. That argument covers the **server**, but not the send loop's catch-up mirror, which `accumulateSentDict:2840` refuses to overwrite. The rationale is incomplete as written.

**Fix:** anchor the orphan commit to the coverage actually achieved (track the highest published chunk end and pass it to `sendCommitMessage`), and add a mid-chunk fault seam plus a test for the partial shape.

### 4. OSS e2e assertion cannot fail for its stated reason — `[OSS] QwpSymbolDictRecycleE2ETest.java:299-306` (in-diff)

> **Problem:** `awaitAckedFsn(preRecycleFsn)` returns true whether or not the epoch base rolled.
> **Net impact:** Loss of regression detection; no user-facing effect.
> **Evidence:** Static, reading `awaitAckedFsn` at client `dd3e9046`:1358-1363.

`preRecycleFsn` is the sender's *first* flush, so it is 0. With a correct base, `internalTarget < 0` short-circuits true. With a broken base of 0, `internalTarget == 0` and the check becomes `ackedFsn() >= 0`, also true — the last batch was awaited acked two statements earlier. No path returns false. The property is covered elsewhere (`SymbolDictRecycleFsnContinuityTest:84` and `:393`), so this is redundancy loss rather than a hole.

**Fix:** take `preRecycleFsn` from a later batch so it is non-zero, or assert epoch advance directly.

## Minor

### 5. Release note overstates a public-API break — PR body

> **Problem:** Release note says `connect(...)` gained parameters in place; it did not.
> **Net impact:** Readers audit a public factory method that never broke.
> **Evidence:** Signature-set diff `981bdb02a`→`dd3e9046`: 11→12 overloads, zero removed.

The note bundles two claims. `CursorWebSocketSendLoop`'s public ctor **did** change in place (7→7 with one base signature gone). `QwpWebSocketSender.connect(...)` is **purely additive** — all 11 base signatures survive byte-identical, one overload added.

## Coverage map

**Test gate: PASSES.** Zero admitted Critical coverage gaps. Four admitted Moderate gaps, each fixable on an existing fixture:

| Gap | Mutation that ships green |
|---|---|
| Arm-recency short-circuit `:5545` | Delete the guard — every armed, undrained `table()` gains a 2 s stall; all 7 starvation tests sleep past the window first |
| `awaitDeferredEngineClose` interrupt policy `:5334`/`:5340` | Delete either restore — zero `interrupt` hits in the deferred-close/slot-heal tests (author-disclosed) |
| Starvation wait's throw-arm interrupt restore | No test sets the interrupt flag before the latch |
| CLOSE_LOOP resume `else` branch `:5827` | All four resume tests force `cap=4096`, so the branch never runs |

The strict SQL-assertion checklist is **clean** on every added/changed ENT and OSS line: zero banned `assertSql`/`assertPlanNoLeakCheck`/`getPlan`, zero `.returnsOnce`. The six new `assertQuery(...).noLeakCheck()` calls all sit inside multi-statement `assertMemoryLeak` scopes, so the anti-pattern rule correctly does not fire.

## Summary

**Verdicts — ENT #1171: approve. OSS #7525: approve with comments (item 4). Client #91: approve with comments (items 1–3).**

Both gates pass: no admitted Critical findings, and no admitted Critical coverage gap.

- **Provenance:** `questdb` OFF-DEFAULT — in scope; `java-questdb-client` OFF-DEFAULT — in scope.
- **Admitted split:** 5 in-diff, 0 out-of-diff-breakage. That is structural, not luck: `questdb-client` is `<scope>test</scope>` in both `questdb-ent/pom.xml` and `questdb/core/pom.xml`, and `io.questdb.client` has zero imports in either repo's production tree (positive control on `questdb/utils` returns hits). No enterprise or OSS production code can be broken by these API changes.
- **Severity distribution:** 4 Moderate, 1 Minor, 0 Critical.
- **Java 8 floor** (`core/pom.xml` pins `1.8`) holds — no Java 9+ constructs in the production diff. No committed binaries, no new TODO/FIXME.
- **The store-and-forward contract holds where it matters most.** `hasInitialConnectRun` latches at the tail of `ensureConnected()` on every branch, and the recycle requires `connected`, so step 7 always takes the ASYNC branch — no producer-thread foreground connect, no reconnect budget on the recycle path.
- **ENT #1171 specifically:** no false-green path in `SqlFailoverQwpRecycleLosslessTest`. `DEDUP UPSERT KEYS(ts)` with a unique per-row `ts` makes the dedup key injective, so replays collapse without collapsing real rows; losses fail the count and fabrications fail the equality. Writer-thread throwables are captured and rethrown on the main thread, and every wait helper polls them. It uses no producer-visible role errors as write-gate evidence — it asserts the opposite.

Three notes that are **not** findings, for context: `questdb-ent/e2e/tests/test_sf_producer_kill9_cycles.py` exists but has no recycle coverage — assessed as an accepted gap, since the client's five JVM crash windows enumerate the recycle's own sequence and the residual delta versus a real `kill -9` is pre-existing SF torn-write territory. Submodule pins on feature branches are merge mechanics, not code defects. And several candidates were omitted for failing the admission gate, including a 30-second producer-park claim whose trigger class the project's own SF checklist explicitly sanctions as a durability-failure, not a transport, stall.

https://claude.ai/code/session_01RZRYHBwMqsSqp3ytLheBQM

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 429 / 493 (87.02%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|77|92|83.70%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|328|377|87.00%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|1|1|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|18|18|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 493 / 571 (86.34%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|77|92|83.70%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|345|404|85.40%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|39|43|90.70%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 502 / 571 (87.92%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|77|92|83.70%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|354|404|87.62%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|39|43|90.70%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

jovfer:
## Tandem code review — level 3

Reviewed as one logical change across all three repos: enterprise questdb/questdb-enterprise#1171, OSS questdb/questdb#7525, client questdb/java-questdb-client#91. The same comment is posted on each.

**Verdict — ENT: approve with comments. OSS: approve with comments. client: approve with comments.**

Correctness gate: PASS (no Critical findings). Test gate: PASS (one Moderate coverage gap, no Critical gap). Severity distribution: 0 Critical / 3 Moderate / 1 Minor, all in-diff, no out-of-diff breakage.

**Submodule provenance** — `questdb` `fef8261 → 8accaa9`: OFF-DEFAULT, in scope. `java-questdb-client` `0b9b576 → 0d81211`: OFF-DEFAULT, in scope. No UPSTREAM-SYNC pointer; the OSS branch's two `origin/master` merges fall outside the three-dot diff.

---

## Moderate

### 1. [client] Public javadoc says `table(...)` is the only place a recycle runs; the resume path also completes one from `flush()`/`at()`/`drain()`

- **Problem:** Two public javadocs name `table(...)` as the recycle's only entry point.
- **Net impact:** Handler authors and API readers get a wrong callback-entry list.
- **Evidence:** `Sender.java:697-706` and `SenderErrorHandler.java:36-42` vs. `QwpWebSocketSender.java:4643`, at client head `0d812116`.

`Sender.resetSymbolDictionary()` states "`table(...)` is the only trigger point: a caller that never starts another row never recycles." `SenderErrorHandler`'s threading section states a rebuild-time quarantine is dispatched synchronously "during a symbol-dictionary recycle rebuild (inside `table(...)`, on the producer thread)".

Both hold only for the *initial* trigger. `ensureConnected()` calls `resumeRecycleIfPending()` at `QwpWebSocketSender.java:4643`, before its `if (connected) return;`, and the `REBUILD` arm runs `completeRecycleRebuild(...)` → `rebuildEngineOrAbandon(...)` → `engineRebuildFactory.rebuild(userErrorHandler())`. `ensureConnected()` is reached from `sendRow()` (so `at()`/`atNow()`) and from `flush()`/`flushAndGetSequence()`. A rebuild-time quarantine can therefore be delivered synchronously on the producer thread from `at()`, `flush()` or `drain()`, not only `table(...)`.

The code is right, and this is already documented correctly elsewhere: `armIfEligible()`'s javadoc (`QwpWebSocketSender.java:5257-5260`) says "A pending resume ... may still complete an already-triggered recycle from a later `flush()`-driven `ensureConnected()` call, with no fresh `table()` call involved." Only the two public-facing docs disagree.

**Suggested fix:** qualify both javadocs to the initial trigger and name the resume path, matching `armIfEligible()`'s wording.

### 2. [client] The deferred-close park's interrupt policy has no test, and the PR body claims it does

- **Problem:** PR body asserts interrupt-policy coverage for both park loops; only one has it.
- **Net impact:** The permanent squash record overstates coverage of an untested policy.
- **Evidence:** `grep -c interrupt SymbolDictRecycleDeferredCloseTest.java` → `0`; `SymbolDictRecycleStarvationTest.java:465,521` are the only two interrupt tests.

The **Tests** paragraph of questdb/java-questdb-client#91 says "Tests pin these: ... the interrupt policy under both park loops". Only `maybeBlockForStarvedReset()`'s policy is pinned, by `testInterruptDuringWaitDoesNotManufactureAbandon` and `testInterruptedWaitTimesOutAndRestoresFlag`. `awaitDeferredEngineClose()` (`QwpWebSocketSender.java:5298-5345`) has none. The **Design decisions** section states the opposite and is the accurate one: "The deferred-close park's interrupt policy is code-reviewed but intentionally untested this round."

Classified Moderate rather than Critical: both `catch (Error)` / `catch (RuntimeException)` arms were read and are a faithful copy of the tested sibling, and the trigger needs a wedged SF worker *and* a producer-thread interrupt. But the observation seam already exists — `deferredCloseParkWitness` is used by `SymbolDictRecycleDeferredCloseTest` and `SymbolDictRecycleSlotHealTest` — and the sibling test shape is directly reusable, so "disproportionate to test" does not apply.

**Suggested fix:** add the test (seam and pattern both already exist), or correct the Tests paragraph so the squashed message does not claim coverage that is not there.

### 3. [ENT] The failover symbol oracle lacks the positive control its OSS twin received on this branch

- **Problem:** ENT mismatch-count oracle has no partitioning positive control.
- **Net impact:** A future SQL-layer change would silently neuter the ENT cross-epoch oracle.
- **Evidence:** `SqlFailoverQwpRecycleLosslessTest.java:498-520` vs. `QwpSymbolDictRecycleE2ETest.java:333-352`; control added by `a3a6c46f95` on this branch.

`assertSymbolsSurvivedFailover` asserts `count(*) WHERE sym IS NULL OR sym <> concat('s', id % 32)` equals `0`, then `count_distinct(sym) == 32`. The second query proves the column is populated but does not partition: if the mismatch predicate ever resolved always-false, a row carrying the wrong symbol would still leave 32 distinct values and both assertions would pass.

The OSS twin of this exact oracle carries the missing half, added on this same branch by `a3a6c46f95` ("Fix round 1: close vacuous-pass gaps in recycle e2e oracles") with precisely this rationale: `count(*) WHERE sym = concat(...)` must equal `TOTAL_ROWS`, so the two queries together account for every row exactly once. The ENT file did not get it.

No defect today — `ConcatFunctionFactory` resolves `concat('s', id % 32)` at this head, so the predicate is live. This is bounded, future-contingent exposure in new test code.

**Suggested fix:** add the matching `WHERE sym = concat(...)` count-equals-total query, as the OSS twin has.

---

## Minor

### 4. [OSS] `awaitAckedFsn` comment describes a failure mode the preceding assertion pre-empts

- **Problem:** Comment credits an assertion with coverage the line above already owns.
- **Net impact:** Reader over-estimates what the last assertion can catch.
- **Evidence:** `QwpSymbolDictRecycleE2ETest.java:303-317` with `awaitAckedFsn` at `QwpWebSocketSender.java:1373-1377`.

Line `:306` asserts `finalEpochBase > lastOrganicFsn`, which makes `internalTarget = targetFsn - fsnEpochBase` strictly negative, so `awaitAckedFsn` takes its prior-epoch short-circuit and returns `true` unconditionally. The comment claims "any lost roll lands it at or above the final epoch's raw ack watermark ... and the await fails" — but a lost roll fails `:306` first, so `:317` never evaluates in that scenario. The assertion is not vacuous (it does pin the negative-target short-circuit); only the comment over-claims.

---

## Coverage map

**Test gate: PASS.** Admitted coverage gaps: 1 (finding 2, Moderate).

Every other behavioural change carries a named test, located by recorded search over `core/src/test/java`: arming/floor → `SymbolDictRecycleArmingTest`; starvation wait and its interrupt policy → `SymbolDictRecycleStarvationTest`; `CLOSE_LOOP`/`REBUILD` resume → `SymbolDictRecycleTest`, `SymbolDictRecycleCrashWindowsTest`; deferred close → `SymbolDictRecycleDeferredCloseTest`, `SymbolDictRecycleSlotHealTest`; breach latch → `SymbolDictRecycleFsnContinuityTest`, `SymbolDictRecycleRefusalTest`; FSN epoch base → `SymbolDictRecycleFsnContinuityTest`; counters → `CursorSendCountersTest`; `markEverConnected` → `SymbolDictRecycleStep7FaultTest`, `SymbolDictRecycleOutageTest`; the `resumeCommitFaultForTesting` seam → `SymbolDictRecycleTest:837,890`; defaults → `SymbolDictResetDefaultsTest`; pooled forwarding → `SenderPoolSfTest`.

## Checked and clean

Recorded so the ground is not re-walked:

- **Store-and-forward contract holds.** `reconnectMaxDurationMillis` reaches only the SYNC *initial* connect: `hasInitialConnectRun` latches one statement after `connected = true`, and `maybeRecycleForDictReset` requires `connected`, so `connected ⟹ hasInitialConnectRun` unconditionally — including `lazy_connect=true` and an ASYNC sender whose socket never came up. The live loop's `connectLoop` retries indefinitely under its named Invariant B. The rebuilt loop is seeded via `markEverConnected()` before `start()`, so a post-recycle endpoint-policy rejection keeps its post-init retry classification instead of latching terminal. `latchRecycleBreach` is the sole `recycleFailure` writer and fires only on a recovered-unacked rebuild.
- **FSN monotonicity** holds across the roll in both accessor read orders; `rollFsnEpochBase`'s `cursorSendLoop == null` precondition cannot be violated, since `ensureConnected` resumes before it can assign a loop.
- **Concurrency: zero candidates** after a full writer/reader inventory of every non-volatile field written inside the swap, plus the listener-detach window, `isRingDrained()`'s read order, and the orphan drainer's `0L` FSN scale.
- **Resources:** no per-recycle thread, fd, mmap or disk-residue leak across epochs; `close()` during `CLOSE_LOOP`/`REBUILD` is covered; pooled give-back correctly carries a pending `REBUILD` to the next borrower.
- **Cap raise** to 2M matches the server constant already at 2M on OSS master; the OSS ordering test pins client ≤ server.
- **Static gates clean** across all three diffs: no banned query-result helpers on added test lines, no `.returnsOnce(...)`, no new TODO/FIXME, all four added LOG chains terminated with no throw-capable arguments.

Four candidates raised during the review did not survive falsification and are not reported: `resetSymbolDictionary()` being inert on `connect(...)`-built senders (net-positive versus base — the cap doubled and the failure message names the remedy); the fuzz test's `restarts/12` disconnect floor (an explicitly reasoned flake margin, non-vacuous); and two resource-management propositions plus one counter-folding proposition that each reduce to an unknown producer with no supported operation reaching them.

https://claude.ai/code/session_01BeJEkMx8gFTcdb1rfVkiUg

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 503 / 571 (88.09%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|77|92|83.70%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|355|404|87.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|39|43|90.70%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 503 / 575 (87.48%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|77|92|83.70%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|355|408|87.01%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|39|43|90.70%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 502 / 566 (88.69%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|78|93|83.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|351|396|88.64%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|41|45|91.11%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

jovfer:
# Tandem code review — level 3

questdb/questdb-enterprise#1171 · questdb/questdb#7525 · questdb/java-questdb-client#91 (branch `qwp-dict-recycle-r1`), reviewed as one logical change.

**Verdict:** client **approve with comments**, OSS **approve**, ENT **approve with comments**. Nothing blocks. The three fixes this round are correct, and CI shows no failure this change causes.

| PR | Head reviewed | Base |
|---|---|---|
| questdb/java-questdb-client#91 | `91fd4451` | `981bdb02` |
| questdb/questdb#7525 | `b8acf7098a` | `6f1d196f57` |
| questdb/questdb-enterprise#1171 | `5532b1590` | `0e4f18001` |

**Scope.** Both submodule pointers move to commits that exist only on the feature branch, so both are in scope:
- `questdb`: OFF-DEFAULT, in scope.
- `java-questdb-client`: OFF-DEFAULT, in scope.

The OSS and ENT test files are byte-identical to the previous review. The only new code is the four client commits. The OSS/ENT merges brought in upstream code only.

**PR metadata.** The titles follow the convention, `Fixes #80` heads the client body, and the labels fit. The body issues are Minor 5.

---

## Moderate

### 1. [client] The public javadocs don't mention the new live-connection gate
- **Problem:** the javadocs describing when the recycle runs, and the monitoring advice, leave out the live-connection gate.
- **Net impact:** API users and operators will misread a recycle that is waiting for an outage to end.
- **Evidence:** compare `Sender.java:698-704` with `QwpWebSocketSender.java:5632-5635`, plus scratch runs at `91fd4451` and `ce20e09b`.

Commit 91fd4451 makes the `table()` barrier return early while `isLinkUp()` is false, before it checks the ring or enters the starvation wait. That is the design the PR body describes. These texts were not updated:
- **`resetSymbolDictionary()`, `Sender.java:698-704`:** says "The reset runs at the next `table(...)` call that finds all published data acknowledged and no row in progress", and blames indefinite deferral only on "sustained saturation".
- **`symbolDictReset`, `Sender.java:1923`:** states the same condition.
- **`symbolDictResetMaxWaitMillis`, `Sender.java:1979-1995`, and the constant comment at `QwpWebSocketSender.java:149-158`:** say the next `table()` "BLOCKS … an outage costs the producer one bounded pause". Now a `table()` call made while the loop is reconnecting skips the wait.
- **Monitoring advice, `Sender.java:2001-2005` and `QwpWebSocketSender.java:2616-2625`:** say that "armed while the epoch does not advance" means no row start saw a drained backlog, or that the one wait already timed out. During an outage the backlog can be drained and the wait is never entered.

Scratch tests, same source at both revisions:

| Trigger | 91fd4451 | ce20e09b |
|---|---|---|
| All data acked, `resetSymbolDictionary()`, server killed, reconnect started, then `table()` | 0 ms, epoch 0, still armed | epoch 1 |
| Armed longer than max wait (500 ms) with unacked rows, server killed, then `table()` | 0 ms, 0 starvation timeouts, no WARN | 500 ms, 1 timeout |

Base has no recycle, so this is a new surface.

**Fix:**
- Add "and the I/O loop holds a live connection" to both trigger sentences.
- In the max-wait javadoc and the constant comment, say that a call made while the loop is reconnecting skips the wait.
- Add "or the sender is between connections" to both monitoring paragraphs.

### 2. [ENT] The failover test's javadoc describes mid-outage swaps the pinned client no longer does
- **Problem:** the test's javadoc says a swap can run mid-outage, and the pinned client no longer allows that.
- **Net impact:** readers will misjudge what the test and its `recycledDuringOutageWindow` log line exercise.
- **Evidence:** compare `SqlFailoverQwpRecycleLosslessTest.java:52-68` with client `QwpWebSocketSender.java:5632-5635` and `CursorWebSocketSendLoop.java:1848`.

What the javadoc says:
- :52-53: "What actually withholds a swap during the outage is `isRingDrained()`".
- :59-61: "if the ring happens to be drained the instant the outage begins, the very next `table()` call runs the swap".
- :63-64: "the swap and the epoch bump return promptly even mid-outage".

What happens at 91fd4451:
- Once the I/O thread sees the drop, `connectLoop` sets `lastReconnectError`, `isLinkUp()` reads false, and the barrier returns before it checks `isRingDrained()`.
- The client's rewritten `SymbolDictRecycleOutageTest` asserts exactly this.
- The same gap applies to :117-120 ("an arm attempt that catches the ring undrained during the outage gives up quickly"): during an observed outage, an arm attempt never reaches the wait.

How it happened: the javadoc was last edited in f522c77e5, when the pin pointed at a client without the gate. d82692f34 moved the pin without touching the javadoc. The ENT PR body now says the javadoc is "written against that path", which isn't true.

No assertion depends on this text. The test passed on the Windows lifecycle leg in both builds.

**Fix:** rewrite :52-68 and :117-120 in terms of the gate. A swap between the demote and the promote can now only land in three places: before the drop is observed, after reconnecting to B, or inside the connect race that earlier reviews accepted. That is what `recycledDuringOutageWindow` measures. Once the javadoc is rewritten, the body sentence becomes true.

### 3. [client] Coverage gap: the fix for the previous review's `build()` leak has no deterministic test
- **Problem:** the closed-only check in `setEngineRebuildFactory` is not pinned by any test.
- **Net impact:** if the check reverts, `build()` throws again, and if the setters also move back outside the guard, the leak returns.
- **Evidence:** with the setter changed back to `checkNotClosed()` at 91fd4451, the whole client suite (3,550 tests) stays green.

Only five test classes call the setter, and all of them use healthy senders. `SenderPoolDataLossNotificationTest` catches the full regression only when a delay is inserted into `build()`:
- Setter reverted and both setters moved back outside the guard: 4/4 red, `MMAP_DEFAULT … 4206592`.
- Setter reverted alone: the pool logs "could not open slot … deferring this and remaining slots", but the test still passes.

A regression would expose the same users as before:
- pool startup recovery where the server rejects the replayed rows as terminal;
- `initial_connect_retry=async` with a 401;
- an SF restart whose backlog gets a terminal NACK.

**Fix:** add a timing-free test that needs no production hook and runs in 0.2 s:
1. Use a raw server that answers every upgrade with 401, and build with `initial_connect_retry=async`.
2. Wait until the error handler receives the TERMINAL error.
3. Check that `table()` throws, which shows the error is latched but the sender is not closed.
4. Call `setEngineRebuildFactory(f)` directly and assert it does not throw.

In a scratch run this test was green 6/6 at head and red 2/2 with the setter reverted.

### 4. [client] Coverage gap: inverting the interrupt assertions removed the only test of the parks' anti-spin clear
- **Problem:** nothing tests any more that each loop iteration in the two recycle parks clears the interrupt flag.
- **Net impact:** if that clear is removed, an interrupted producer spins a core for up to 2 s (starvation wait) or 30 s (deferred close).
- **Evidence:** with the clear deleted in both parks, 79/79 recycle tests are green at 91fd4451. The same edit at ce20e09b turns 2 tests red.

Without `wasInterrupted |= Thread.interrupted()` (`:5337`, `:5597`), `parkNanos` returns immediately while the flag is set, so the loop spins instead of parking.
- The old swallow assertions caught this indirectly.
- The inverted "flag handed back" assertions also pass with the clear deleted, because the flag is simply never cleared.
- The timeout tests don't catch it either, because the loop still respects its deadline.
- Both test javadocs still say the park "cannot busy-spin".

Producer-thread CPU time vs wall time with the clear deleted:
- 150 ms drained wait: head uses 7–8 ms of CPU, the mutant 149–154 ms.
- 300 ms timeout wait: head uses 14–17 ms, the mutant 300 ms.
- Under 2× CPU oversubscription the mutant still used 215 of 300 ms.

**Fix:** in the two timeout-exit interrupt tests, assert that `ThreadMXBean.getCurrentThreadCpuTime()` over the `table()` call stays below half the wall time. Guard it with `Assume` on `isCurrentThreadCpuTimeSupported()`. The API has existed since JDK 5, so it respects the Java 8 floor.

---

## Minor

### 5. [OSS][ENT] The PR bodies predate the gate
- **Problem:** the OSS body leaves the gate out of its list of client changes, and the ENT body says its test javadoc matches the gate when it does not.
- **Net impact:** reviewers read a description that disagrees with the pinned client.
- **Evidence:** the OSS body text, compared with `QwpWebSocketSender.java:5632-5639`.

The OSS body says "the fuzz producer can pause up to 2 s … when a bounce catches the ring undrained". With the gate, the producer pauses only if the wait began while the link was up. For ENT, see finding 2: the body claims the javadoc is "written against that path", and it repeats the "gives up quickly rather than parking the writer" wording.

---

## Coverage map

The test gate passes: there is no Critical coverage gap, and two Moderate gaps (findings 3 and 4). The new gate itself is well tested:
- Deleting the gate, or its `lastReconnectError == null` condition, turns both rewritten outage tests red.
- Reverting either interrupt hand-back turns its inverted assertion red.

## Fix verification

**`build()` leak.** A 300–500 ms delay was forced after connect:
- At head, `build()` returns, the first call surfaces the terminal error, and nothing leaks.
- At ce20e09b, it leaks `MMAP_DEFAULT … 4206592`, the same number as the earlier Windows CI failure.
- The client's Windows CI leg is now green.

**Live-connection gate.**
- With a 2 s token pull that ignores interrupts, ce20e09b's `table()` took 2,001 ms. At head it returns in 0 ms and the recycle stays armed.
- On healthy links the gate read "up" in all 12 combinations of connect mode × memory/SF × durable-ack, and every barrier swapped.
- A stress run made 2.45 billion concurrent `isLinkUp()` calls across 4,145 client swaps, with no exceptions.

**Interrupt hand-back.** With the interrupt flag set on entry, all of these committed and handed the flag back:
- the heal pass;
- a deferred close of a recovered engine;
- a memory-mode REBUILD resume;
- 20,000 rows.

**Store-and-forward contract.** Nothing changed this round lets a transport error, a reconnect time limit, or a hard failure reach the producer.

## CI

- **Client:** all 16 checks green, including Windows.
- **ENT PR-merge build 272038:** green.
- **enterprise-ci build 272039:** one red test on `windows-amd64-lifecycle`, `EntLifecycleOrchestratorTest.testSwitchRoleOwnershipRefusalTornArmFuzz`. It timed out after 30 s inside `LOG.advisory → nextBully`. It is not caused by this change:
  - Lines logged within the same millisecond reached the agent 63–65 s later. The log consumer was blocked writing to the agent's stdout, so the log ring filled up.
  - The PR changes no ENT or OSS production code, and the test imports nothing from the client.
  - The same test passed in 281 ms in build 272038.
  - `SqlFailoverQwpRecycleLosslessTest` passed on that same leg.

  Rerun the stage.

## Summary

| PR | Verdict |
|---|---|
| questdb/java-questdb-client#91 | approve with comments (findings 1, 3, 4) |
| questdb/questdb#7525 | approve (Minor 5 only) |
| questdb/questdb-enterprise#1171 | approve with comments (finding 2) |

- **Severity:** 0 Critical, 4 Moderate, 1 Minor, all inside the diff.
- **Test gate:** passes.

bluestreak01 (changes requested):
@jovfer

## PR #91 — level 3

### Critical

**Problem:** A released public constructor no longer links after upgrading.  
**Net impact:** Direct policy-aware loop users cannot construct their ingestion loop.  
**Evidence:** Identical 1.3.8-compiled caller passes release/base `981bdb02`; HEAD `91fd4451` throws `NoSuchMethodError`.

**Location:** `core/src/main/java/io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java:945–964` — **in-diff**.

Adding `externalFsnBase` replaces the published 13-argument constructor instead of preserving it. Applications compiled against that constructor fail when replacing the dependency, before connection or ingestion begins. This is a deterministic, net-negative compatibility regression; facade-only callers are unaffected.

Independent falsification checked whether this was internal or unpublished: the constructor is documented, exported, and present in the published **1.3.8 JAR**. Retained shorter overloads cannot satisfy its JVM descriptor.

**Fix:** Preserve the 13-argument overload, delegating with `externalFsnBase = 0L`, and pin its signature in compatibility tests. That bridge passed the unchanged caller and 135 focused tests in scratch.

[Reproducer]( /tmp/pr91-review-zo1dar/ReleasedLoopCaller.java) · [Base output](/tmp/pr91-review-zo1dar/released-caller-base-tree.log) · [HEAD failure](/tmp/pr91-review-zo1dar/released-caller-head.log)

### Coverage and validation

- **Test gate: pass — 0 admitted coverage gaps.**
- JDK 8: 490 tests across focused runs; zero failures/errors, one skipped.
- JDK 25: core production and test compilation passed.
- Reviewed linked [OSS #7525](https://github.com/questdb/questdb/pull/7525) real-server/fuzz coverage and [Enterprise #1171](https://github.com/questdb/questdb-enterprise/pull/1171) failover coverage, including assertions and CI wiring; not executed locally.

### Provenance

- Client PR #91: no submodule-pointer changes.
- Enterprise → OSS `dd7ad37a`: **OFF-DEFAULT**.
- OSS → client `91fd4451`: **OFF-DEFAULT**.

### Summary

**Verdict: request changes.**  
Severity: **1 Critical, 0 Moderate, 0 Minor**. Attribution: **1 in-diff, 0 out-of-diff-breakage**.

Primary checkout unchanged; no review posted.

jovfer:
The released 13-argument `CursorWebSocketSendLoop` constructor is back as an overload delegating with `externalFsnBase = 0`, and `ExportedApiCompatibilityTest` now pins all seven public constructors present in the 1.3.9 jar, copied from its `javap` output. An exported-signature diff of this head against the merge base (`javap` over every exported package, at the merge base, at the head and in the 1.3.9 jar) found one more in-place widening, `QwpWebSocketSender.connectWithCredentialSupplier` (23 → 26 arguments; on main since the OIDC device-flow change, not in any release); its 23-argument form is back the same way and both merge-base overloads are pinned. After both, the diff against the merge base is empty; against the 1.3.9 jar it lists 18 synthetic `Outer$1` anonymous-class artifacts that were already absent at the merge base. Head: ba54a138.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 502 / 569 (88.22%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|78|93|83.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|41|47|87.23%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|351|397|88.41%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 502 / 569 (88.22%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|78|93|83.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|41|47|87.23%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|351|397|88.41%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 503 / 569 (88.40%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/Sender.java|78|93|83.87%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorWebSocketSendLoop.java|41|47|87.23%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/QwpWebSocketSender.java|352|397|88.66%|
|:large_blue_circle:|io/questdb/client/impl/PooledSender.java|2|2|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendEngine.java|11|11|100.00%|
|:large_blue_circle:|io/questdb/client/impl/ConfigSchema.java|3|3|100.00%|
|:large_blue_circle:|io/questdb/client/cutlass/qwp/client/sf/cursor/CursorSendCounters.java|16|16|100.00%|

bluestreak01 (changes requested):
@jovfer

## PR #91 review, level 3

**Verdict: request changes.** One Critical finding, one Moderate coverage gap, no Minor.

- **Revisions:** PR head `aaabff3e`, merge base `916e51f6`.
- **Submodules:** the PR moves no submodule pointers.
- **Title and description:** the title follows Conventional Commits, `Fixes #80` is at the top of the body, and the labels fit.

### Critical

**Problem:** A recycle unlocks the slot, so a second sender can take it.
**Net impact:** The running sender with the duplicate `sender_id` throws on every call while the intruder lives.
**Evidence:** A probe run on JDK 8. Head `aaabff3e` reproduced it in 2 of 2 runs. Base `916e51f6` did not reproduce it: 583,500 of 583,500 attempts by B were refused.

**Where:** this is in-diff. The chain is:
1. `QwpWebSocketSender.java:5871-5872`: step 3 calls `outgoing.close()`. This releases the slot's directory lock and removes the logical lock file.
2. `:5481`: `rebuildEngineOrAbandon`.
3. `:5726`: `engineRebuildFactory.rebuild`.
4. `Sender.java:3449`: `constructEngineOnSlot` takes the locks again.

**The contract this breaks:** the `senderId` javadoc at `Sender.java:2987-2996` says the slot is "locked exclusively for the sender's lifetime via `flock`" and that "the second sender to start with a colliding id fails fast with 'sf slot already in use'".

**Trigger:**
- Sender A is running with the default `symbol_dict_reset=on` and `sf_dir` set.
- Sender B is configured with the same `sf_dir` and `sender_id`, and retries `build()` after it fails. Realistic sources of B are a rolling-deploy overlap, a supervisor restart loop, or an accidental duplicate instance.
- B's `build()` lands between A's step-3 close and its step-4 rebuild, and succeeds.
- A's rebuild then fails with `SlotLockContentionException` and stays in the `REBUILD` resume state with `cursorEngine == null`.
- Every later `table()`, `at()` or `flush()` on A throws "symbol dictionary recycle could not rebuild its engine; retried on the next send". The message says the failure is transient, but it lasts as long as B holds the slot.

**Probe results:** the probe is a scratch JUnit test. Sender A ingests and flushes in a loop while a second thread calls `Sender.fromConfig(cfg)` for B in a tight loop, with the same `sf_dir` and `sender_id` at both revisions.

| Revision | B's `build()` | A while B is open | A after B closes |
|---|---|---|---|
| Head, run 1 | succeeded on attempt 95,639, when A was at epoch 1 | 0 rows ok, 2,630 exceptions | 50 of 50 ok |
| Head, run 2 | succeeded when A was at epoch 3 | 0 rows ok, 2,599 exceptions | 50 of 50 ok |
| Base | 0 of 583,500 attempts succeeded, all `SlotLockContentionException` | 2,057,013 rows ok, 0 errors | – |

The probe triggers recycles with `resetSymbolDictionary()` to raise their frequency. Even so, B won within the first few completed recycles, so the window is the whole close-and-rebuild span, not a few microseconds. A natural threshold recycle opens the same window.

**Net effect:** this is worse than base (net-negative).
- **Who is affected:** deployments where two senders share `sf_dir` and `sender_id` at the same time. The product explicitly anticipates this case and documents the outcome.
- **Change from base:** at base the newcomer fails fast and the running producer carries on. At head the running producer stops accepting rows for an unbounded time, and its error message is misleading.
- **Other safeguards:** there is no silent loss, because the barrier proved everything acknowledged and new rows are rejected loudly. Nothing else found holds the slot during the window; I searched `SlotLock`, `constructEngineOnSlot` and the recycle steps.
- **Store-and-forward rule:** this also breaks the rule that a running store-and-forward producer never hard-fails after initialization.

**Suggested fix:** keep the slot exclusively owned across steps 3 and 4.
1. Before step 3, have the rebuild factory take `SlotLock.acquireLogical(slotPath)`.
2. Close the outgoing engine with `close(false)`, which already exists for callers that hold the logical lock and skips removing the lock file.
3. Rebuild through `constructEngineOnSlotLocked` while the lock is still held, so a colliding `build()` fails when it tries to take the logical lock.
4. Add a regression test shaped like the probe, and check that it is red on the current head.

### Moderate

**Problem:** The per-window starvation resets and the age guard are untested.
**Net impact:** A regression would pause each `table()` right after arming, or stop recycling after the first starvation wait. Either way the harm is bounded.
**Evidence:** mutation runs at `aaabff3e`. Under each mutation below, all 148 tests matching `SymbolDictRecycle*Test`, `SymbolDictResetDefaultsTest` and `SenderPoolSfTest` stay green.
- M1: delete `QwpWebSocketSender.java:5348` (`starvationWaitDoneThisArm = false`).
- M2: delete `:5347` (`armedSinceNanos = …`).
- M3: delete the age guard at `:5644`.

Why nothing catches them:
- No test arms a second window after a starvation wait has already run.
- Every test that enters the wait first sleeps for `maxWait + 50` ms.
- `testContinuousFlushesPreserveStarvationWindow`, added in this PR's latest commit, only asserts the final count of 1.

**Suggested fix:** extend that test to cover a second armed window.
1. Release acks until the first recycle commits (epoch 1).
2. Gate acks again and re-arm with `resetSymbolDictionary()`.
3. Assert that a `table()` inside `maxWait` returns promptly with the timeout count unchanged. This catches M2 and M3.
4. Assert that the count then reaches 2. This catches M1.

### Coverage map
- **Test gate:** passes. The only admitted coverage gap is the Moderate one above.
- **Local run:** 484 related tests on JDK 8 at head, with 0 failures, 0 errors and 1 skipped.
- **Tandem PRs:** both exist and are cross-linked with this PR: OSS questdb/questdb#7525 (e2e and fuzz) and Enterprise questdb/questdb-enterprise#1171 (failover). #7525's submodule pin points at `ba54a138`; head adds only javadoc and test commits on top of it.

### Summary
- **Severity:** 1 Critical, 1 Moderate, 0 Minor. Both are in-diff; no out-of-diff breakage was admitted.
- **Checked and not reported:** lost or duplicated rows, epoch-offset errors in the public sequence numbers (FSNs), symbol IDs referenced before registration, durable-ack mode, deferred-commit groups, and binary compatibility.

PR #97, opened 2026-09-11
fix(build): stabilize the token-store rename retry test

mtopolnik:
## Summary

Fix a timing-sensitive token-store test that failed in the `mac-other` job in [QuestDB CI build 269255](https://dev.azure.com/questdb/questdb/_build/results?buildId=269255), reported in questdb/questdb#7263.

The test checks that replacing a file succeeds after a permission error. Previously, a background thread had about 80 ms to restore write permission. If that thread ran too late, the test failed even though the retry logic worked correctly.

The test now restores permission immediately after the first failed attempt, without a background thread. It checks that the operation retries exactly once, writes the expected contents, and leaves no temporary file. It also restores the original permissions even if the test fails.

Production behavior stays the same: retry limits, delays, and interrupt handling do not change. The implementation adds a private callback for the test and a null check on retries. The test still needs POSIX file permissions and skips execution as root. It does not test Windows file-sharing errors.

## Validation

- All 90 `FileTokenStoreTest` tests pass through Maven on macOS arm64 with Java 8 and Java 25.
- Applying the fix to current `main` adds no other Java changes. The only intervening commit changes the PR-title CI check.
- `git diff --check` passes.

### DIFF
--- a/core/src/main/java/io/questdb/client/cutlass/auth/FileTokenStore.java
+++ b/core/src/main/java/io/questdb/client/cutlass/auth/FileTokenStore.java
@@ -987,6 +987,10 @@ public final class FileTokenStore implements TokenStore {
 
 
     private static void replaceTarget(Path tmp, Path target) throws IOException {
+        replaceTarget(tmp, target, null);
+    }
+
+    private static void replaceTarget(Path tmp, Path target, Runnable beforeRetryForTesting) throws IOException {
         // atomically rename tmp over target. On Windows a concurrent reader in any process holding target open
         // can make the rename fail transiently with AccessDeniedException (a sharing violation); retry a few
         // times on a short backoff before giving up, so a routine read/write overlap does not needlessly degrade
@@ -1009,6 +1013,10 @@ public final class FileTokenStore implements TokenStore {
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
                     break;
+                }
+                // Tests clear a real permission denial before the next move attempt.
+                if (beforeRetryForTesting != null) {
+                    beforeRetryForTesting.run();
                 }
             }
             try {
--- a/core/src/test/java/io/questdb/client/test/cutlass/auth/FileTokenStoreTest.java
+++ b/core/src/test/java/io/questdb/client/test/cutlass/auth/FileTokenStoreTest.java
@@ -73,8 +73,7 @@ import static io.questdb.client.test.tools.TestUtils.repeat;
 /**
  * Coverage for {@link FileTokenStore}.
  * <p>
- * PLATFORM SCOPE. CI runs Linux only, so the store's Windows-motivated arms are covered here to the extent a
- * POSIX host can reach them, and no further:
+ * PLATFORM SCOPE. POSIX hosts cover the store's Windows-motivated arms only to the extent described below:
  * <ul>
  *     <li>the {@code AccessDeniedException} retry in {@code replaceTarget} - the sharing violation a Windows
  *     reader holding the target open produces - IS exercised, by denying the rename with directory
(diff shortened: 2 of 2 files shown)


mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 4 / 4 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/auth/FileTokenStore.java|4|4|100.00%|

mtopolnik:
## Review: PR #97 at `e0243e31`

No findings: 0 Critical, 0 Moderate, 0 Minor.

Production callers retain retry limits, interrupt handling, and cleanup behavior. The private callback adds a test seam without changing public contracts.

### Coverage
- Test gate: **passes**; 0 admitted coverage gaps.
- `FileTokenStoreTest`: all 90 tests passed on Java 8 and Java 25, with no skips.
- `git diff --check` passed.
- No server/HA tandem required.
- Submodule provenance: no pointer changes.

### Summary
**Verdict: approve.**

Windows sharing-violation coverage remains outside this POSIX test’s scope.

mtopolnik:
#### [PR Coverage check]

:heart_eyes: **pass** : 4 / 4 (100.00%)




#### file detail

|----|----|----|----|----|
|:large_blue_circle:|io/questdb/client/cutlass/auth/FileTokenStore.java|4|4|100.00%|

bluestreak01 (approved):
## Review of PR #97 at level 3 (head `e0243e31`, base `981bdb02`)

I found nothing blocking or worth fixing: no Critical, Moderate or Minor findings.

**What changed.** The test now clears the permission denial through a private hook that runs on the calling thread. It replaces a background thread that had about 80 ms to do the same job. The only production caller, `FileTokenStore.save()` (L639), still uses the 2-argument `replaceTarget`, which passes `null`. For production, the retry count, the 20 ms backoff, the interrupt handling and the rethrow of the last denial all stay the same. The hook runs only after the sleep finishes normally, so an interrupt still exits the loop before the hook would run.

**PR metadata.** The title uses Conventional Commits. `fix(build)` is a slightly loose label for a change that only touches a test, but the repo's PR-title check accepts it, so I'm not raising it. There's no `Fixes #` line, which is correct because no issue exists; the failure was reported on questdb/questdb#7263, a pull request.

### Evidence I ran
All runs were in a scratch worktree at `e0243e31`:
- **Full test class:** all 90 `FileTokenStoreTest` tests pass on JDK 8 (corretto 1.8.0_492) and JDK 25, run as a non-root user, with none skipped.
- **The old flake mechanism is real.** At base, I raised the old background thread's 30 ms delay to 150 ms and to 400 ms to simulate a late thread. The test then failed with `AccessDeniedException` because the retries ran out.
- **The new test is stable under load.** The three `replaceTarget` tests passed three times in a row with every CPU core saturated at twice the core count.
- **The new test catches real breakage.** Each of these deliberate breakages makes it fail:
  - only one attempt allowed (`REPLACE_MAX_ATTEMPTS=1`): `AccessDeniedException`;
  - returning after the first denial instead of retrying: assertion failure;
  - dropping the hook call: `AccessDeniedException`;
  - removing the `r-x` permission change: `expected:<1> but was:<0>`.

  The last one matters because the PR removed the old up-front check that the denial really happens. The new `retries == 1` assertion covers that case: it fails if the denial never takes effect on the host.
- **The other two `replaceTarget` tests still pass.** They look up the 2-argument method by reflection, and it still exists.
- **Java 8:** only `Runnable`, a lambda and `AtomicInteger` are used. No test imports were left unused.

### Coverage map
- **Test gate:** passes, with 0 coverage gaps.
- **Testing in other repos:** not required. This is local file handling that unit tests can cover, with no server, wire-protocol or high-availability effect. I searched for a same-named branch in `questdb/questdb` and `questdb/questdb-enterprise` and found none, which is expected.

### Summary
- **Verdict: approve.**
- **Submodules:** no submodule pointers changed.
- **Findings:** none, either in the diff or at unchanged callers. The only out-of-diff callers are `save()` and the two other `replaceTarget` tests, and I checked all three.
- **Remaining limits:**
  - The test still needs POSIX permissions and skips when run as root.
  - A real Windows sharing violation is still not tested, as the class javadoc says.
  - Production code now carries a private parameter used only by the test. It is `null` in production and only checked on the retry path after a denial.

